PR #10, opened 2015-04-17
Strip off the leading "1 " from the returned responses

hellais:
@salarcon215 @technosopher Is this actually needed? Can't the data format be how it was before?

### DIFF
--- a/update-bouncer.py
+++ b/update-bouncer.py
@@ -27,12 +27,14 @@ def read_parts_from_mlabns():
     part_list = []
     for sliver in sliver_list:
         try:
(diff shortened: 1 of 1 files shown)


hellais:
@salarcon215 @technosopher can you confirm this is now fixed in the MLAB NS? If so I am going to close this issue.

Thanks.

salarcon215:
Sure, asking for an answer from the team now.

salarcon215:
@mtlynch confirmed that this workaround is no longer needed and can be closed. The root cause was fixed per this:
https://github.com/m-lab/mlab-ns/commit/869581bfdef657518ca46f0d8a6845fe746aeb70
Thanks!

PR #11, opened 2015-07-24
Add ansible tasks to setup and control docker SYSv init scripts

anadahz:
Tasks:
* Setup docker SYSV init.d script
* Enable docker SYSV init.d script
* Start docker SYSV init.d script
* Restart docker SYSV init.d script
* Stop docker SYSV init.d script

Template:
Add docker SYSV init.d script

### DIFF
--- a/ansible/roles/docker/tasks/install-docker-sysv-init-script.yml
+++ b/ansible/roles/docker/tasks/install-docker-sysv-init-script.yml
@@ -1,0 +1,6 @@
+---
+- name: Add docker SYSV init.d script "{{ docker_name_container }}"
+  template: src=docker-container.sysv.j2 group=root mode=0751 owner=root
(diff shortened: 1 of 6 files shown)


hellais:
I think we are going to proceed in the meantime to a non docker solution since on non bare metal machines it's easier to just spin up a new VM than it is to run docker.

I am nonetheless not going to close this as it's very useful stuff. Thanks for working on it :+1:

anadahz:
SYSV  init scripts were most useful to older Debian releases (i.e wheezy).
I guess we should close this PR.
@hellais  do we have a git-grave for all these possibly useful commits?

hellais:
I am not sure what a git-grave is, but as long as you don't delete the branch it stays here.

anadahz:
@hellais git-grave, a place potentially useful things can be stored "forever" :)

PR #16, opened 2015-11-18
Start outlining the structure for the deploy task

hellais:
@joelanders I started outlining what needs to be done to automate deployment on ec2.

### DIFF
--- a/ansible/deploy.yaml
+++ b/ansible/deploy.yaml
@@ -4,9 +4,9 @@
   connection: local
   gather_facts: false
   vars:
-    keypair: ooni-pipeline
+    keypair: ooni-backend
     instance_type: m3.medium
-    security_group: ooni-pipeline-computer-1
+    security_group: ooni-backend-1
     image: ami-1ecae776
     region: us-east-1
   tasks:
@@ -22,7 +22,7 @@
       register: ec2
 
     - name: Add new instances to host group
-      add_host: hostname={{item.public_ip}} groupname=computer
+      add_host: hostname={{item.public_ip}} groupname=backend
       with_items: ec2.instances
 
     - name: Wait for the instances to boot by checking the ssh port
@@ -32,11 +32,11 @@
 #   Use the ec2_vol module to create volumes for attachment to each instance.
 #   Use with_items to attach to each instance (by returned id) launched previously.
 
-# 
+#
 #   - name: Create a volume and attach
 #     ec2_vol: volume_size=20 instance={{item.id}}
 #     with_items: ec2.instances
-# 
+#
 
 - name: Configure instance
   hosts: computer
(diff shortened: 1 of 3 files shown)


joelanders:
Got it. So the roles we're configuring correspond to the Dockerfiles currently under the 'ooni-sysadmin/docker' directory: 'backend,' 'bouncer,' and 'probe.'

(The four Dockerfiles under 'data-pipeline,' I presume, are old and should be deleted?)

The 'backend' and 'bouncer' roles are both served by the 'ooni-sysadmin/ansible/install-oonibackend.yml' playbook, with an 'enable_bouncer=yes' variable toggling the difference.

The 'probe' role is served by the 'ooni-sysadmin/ansible/install-ooniprobe-bridge-reachability.yml' playbook.

So the big change here is that we're taking out the Docker indirection. The only configuration we'll have to move out of the Dockerfiles are the volume mappings. When we code the configuration for the S3 buckets, we might as well just mount them where each application expects them.

^All that's going to happen after we have some EC2 instance(s) running.

What should the top level (Invoke) commands be like? Three separate `start_{backend,bouncer,probe}` commands which ensure an instance is running with a `ooni_role={backend,bouncer,probe}` EC2 tag, then apply the Ansible role to that instance?

Talk to ya in the mornin'.

hellais:
> Got it. So the roles we're configuring correspond to the Dockerfiles currently under the 'ooni-sysadmin/docker' directory: 'backend,' 'bouncer,' and 'probe.'

Correct.

> (The four Dockerfiles under 'data-pipeline,' I presume, are old and should be deleted?)

This is true.

> The 'backend' and 'bouncer' roles are both served by the 'ooni-sysadmin/ansible/install-oonibackend.yml' playbook, with an 'enable_bouncer=yes' variable toggling the difference.

Don't forget to remove this from the sub-repository and make it into the main repository. @mmaker has done this already and perhaps can give some feedback on the process.

> The 'probe' role is served by the 'ooni-sysadmin/ansible/install-ooniprobe-bridge-reachability.yml' playbook.
> So the big change here is that we're taking out the Docker indirection. The only configuration we'll have to move out of the Dockerfiles are the volume mappings. When we code the configuration for the S3 buckets, we might as well just mount them where each application expects them.
> ^All that's going to happen after we have some EC2 instance(s) running.

This is all accurate.

> What should the top level (Invoke) commands be like? Three separate start_{backend,bouncer,probe} commands which ensure an instance is running with a ooni_role={backend,bouncer,probe} EC2 tag, then apply the Ansible role to that instance?

I think this a very good strategy, though I would pass it as an argument to the deploy task as I outlined in this commit.
This means you would run:

`invoke deploy --component $COMPONENT_NAME`

Don't forget that we also need the components **api**, **pipeline**.

I would start out with the API one since that is currently not running and it's blocking on some other projects.

Keep typing 'em keys as well as you know to :+1: 

> Talk to ya in the mornin'.

Indeed :smile:

hellais:
We don't rely on ec2 anymore. Closing this.

PR #17, opened 2015-11-25
replace ansible with a very small shell script

joelanders:
`sudo sh install-ooni-backend.sh` on a fresh debian jessie ec2 instance gets `oonib` running as a service under systemd.

(some things i should polish, but i just tested it on a new instance and it works)


hellais:
This will eventually have to become ansibilised and piloted from the invoke task runner.

Ansible could also run the shell script, though it's probably best to have the control of it be in the hands of ansible.

For the time being since we are going for MVPs I am going to merge this as is since it's clean and works.

Thanks! :+1:

PR #29, opened 2015-12-11
Fix/ansible ooniprobe role

anadahz:


ansible/roles/ooniprobe/tasks/install-git.yml
```diff
@@ -34,4 +13,4 @@
     name={{ item }}
   with_items:
     - cffi
-    - ooniprobe
+    - 'git+https://github.com/TheTorProject/ooni-probe.git#egg=ooniprobe'
```
hellais: Let's be careful with this. Since it leads to the installation of a potentially unstable version of ooniprobe. I would make it a switch to install the master branch version vs installing it from the pypi release.

anadahz: This is ooniprobe git version install task, for the stable ooniprobe there is another [task](https://github.com/TheTorProject/ooni-sysadmin/blob/master/ansible/roles/ooniprobe/tasks/install.yml)  which install ooniprobe package from Debian stretch repository.

PR #36, opened 2016-04-12
Point to new location for pip

hellais:


SuperQ:
Why not use an apt package install of python-pip?

hellais:
Because the dockerfile that is using this ansible role is based on debian wheezy, that has an older version of pip (1.1) that doesn't do certificate validation when fetching HTTPS dependencies :(.

I guess that since the latest stable of debian now ships with a decent version of pip, we could update the Dockerfile and use python-pip instead.

SuperQ:
:crying_cat_face:

anadahz:
This has been fixed in #38

anadahz:
This is still needed my bad.
Merging!

PR #40, opened 2016-04-14
Add support for non virtualenv backend, fix dependencies

anadahz:
Adds supports for installing ooni-backend outside a virtualenv instance.
Fixes pip dependency issues by installing the packages with the correct
order and removes some conflicting python system (apt) packages.


hellais:
LGTM.

PR #46, opened 2016-04-28
Fix/docker volumes dirs

anadahz:


docker/backend/start-collector.sh
```diff
@@ -1,4 +1,3 @@
 #!/bin/sh
-mkdir -p /data/collector/archive
-mkdir -p /data/collector/tor
+mkdir -p /data/collector/{archive,tor,decks,inpurs,raw_reports,etc,var/log/ooni}
```
hellais: I think there is a typo here `inpurs`

Issue #56, opened 2016-06-03
Create an ansible letsencrypt role to provision letsencrypt certificates for ooni-backend


anadahz:
This task is about creating a letsnencrypt ansible role to generate and deploy TLS certificates for ooni-backend.

anadahz:
This has been done, tested and deployed.

PR #59, opened 2016-06-07
Feature/ansible/docker letsencrypt role

anadahz:
Thanks to @hellais for the pointers!

### DIFF
--- a/ansible/install-letsencrypt.yml
+++ b/ansible/install-letsencrypt.yml
@@ -2,12 +2,15 @@
 - hosts: local:staging:testing
   vars:
     ansible_python_interpreter: "/usr/bin/python"
+    bind_oonibackend_data_path: "/data/{{ app_env }}"
+    bind_oonibackend_tls_dir: "{{ bind_oonibackend_data_path }}/tls"
   roles:
     - common
-    - docker_py
+    - { role: docker_py, when: docker_deployment is defined }
+    - { role: letsencrypt, when: letsencrypt_docker is undefined }
 
 - hosts: local:staging:testing
   vars:
     ansible_python_interpreter: "/root/venv/bin/python2.7"
   roles:
-    - letsencrypt
+    - { role: letsencrypt, when: letsencrypt_docker is defined }
--- a/ansible/install-letsencrypt.yml
+++ b/ansible/install-letsencrypt.yml
@@ -3,10 +3,11 @@
   vars:
     ansible_python_interpreter: "/usr/bin/python"
   roles:
+    - common
     - docker_py
 
 - hosts: local:custom
   vars:
-    ansible_python_interpreter: "{{ ansible_env.HOME }}/venv/bin/python2.7"
+    ansible_python_interpreter: "/root/venv/bin/python2.7"
   roles:
     - letsencrypt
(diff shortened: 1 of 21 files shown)


anadahz:
This fixes #56.

ansible/install-letsencrypt.yml
```diff
@@ -3,10 +3,11 @@
   vars:
     ansible_python_interpreter: "/usr/bin/python"
   roles:
+    - common
     - docker_py
 
 - hosts: local:custom
   vars:
-    ansible_python_interpreter: "{{ ansible_env.HOME }}/venv/bin/python2.7"
+    ansible_python_interpreter: "/root/venv/bin/python2.7"
```
anadahz: We should find a better way to do this.
In my case (ansible version 2.0.2.0) `ansible_env.HOME` is not expanding correctly.

anadahz:
This is ready to be merged!

ansible/roles/letsencrypt/vars/main.yml
```diff
@@ -0,0 +1,28 @@
+letsencrypt_docker_image: quay.io/letsencrypt/letsencrypt
+letsencrypt_docker_name: certbot
+letsencrypt_admin_email: admin@openobservatory.org
+letsencrypt_workdir: /var/lib/letsencrypt
+letsencrypt_port: 80
+letsecnrypt_pre_hook: docker stop {{ app_env }}
```
hellais: I think there is a typo here.

ansible/roles/letsencrypt/hosts
```diff
@@ -0,0 +1,6 @@
+[custom]
+hostname
+
+[custom:vars]
+# Testing uncomment to generate test certificates
+letsencrypt_test_cert=true
```
hellais: Should the default not be commented out?

hellais:
A part from those two minor comments this diff looks good to me. Is there a recommended way for me to test this?

ansible/roles/letsencrypt/vars/main.yml
anadahz: No this is not a typo. I have fixed hosts file to include this variable.

ansible/roles/letsencrypt/hosts
anadahz: Good catch, thanks!
Fixed hosts files

anadahz:
@hellais thanks for reviewing this!
I have added documentation in https://github.com/TheTorProject/ooni-sysadmin/pull/59/commits/0cf8b783e92cca7fad44e14711c3eee03ff4b552 on how you can test this.

ansible/roles/letsencrypt/vars/main.yml
hellais: shouldn't it read `letsencrypt_pre_hook` instead of `letsecnrypt_pre_hook`?

anadahz: I though you meant `app_env`. 
You are right though technically it's OK since the variable has the same name (consistent typo).
Thank you for spotting this!

anadahz:
https://github.com/TheTorProject/ooni-sysadmin/pull/59/commits/c15815ac9d85ab3ed51c17ff5486c0fd1fd78be3 resolves the problem of restarting ooni-backend docker containers when TLS certificates are renewed, since it's not that possible and decent to restart a docker container inside another container.

hellais:
Looks good to me!

Issue #55, opened 2016-06-03
Improve automation of ansible role for ooni-backend


anadahz:
Ansible role should better automate ooni-backend deployment.

-  [x] Set up a daily job to generate the bouncer.yaml from mlabns pulled data.
-  [x] Create a `bouncer-base.yaml` with the appropriate HS hostnames generated during the setup.
-  [x] Include a configuration option to deploy the bouncer with already given HS hostnames and keys.

anadahz:
This has been done, tested, merged and deployed.

Issue #49, opened 2016-05-13
Update scripts for installing ooni-backend with SSL support


hellais:
The TLS endpoint branch adds a new configuration option called `collector_endpoints` and `bouncer_endpoints`.

We should update the ansible scripts so that they take into account these changes to the configuration files and set them accordingly if we wish to setup a SSL collector and bouncer.

hellais:
Note: the script technically support TLS, but we should remember to not include the TCP endpoint in it when `enable_tls_*` is set.

PR #62, opened 2016-06-20
Improve ooni-backend role automotation

anadahz:
* Add default config options for stage and testing backends
* Template MLab NS bouncer configuration and setup a daily cronjob
* Wait for backend to bootstrap and generate bouncer's configuration
* Ensure that the latest certificate is being used and move to the
correct PATH
* Add support for HTTPS alternate collectors and bouncers


anadahz:
Work in progress.

ansible/roles/ooni-backend/templates/update_bouncer.py.j2
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env python
+
+import yaml
+import json
+import urllib
+
+MLAB_NS_QUERY_URL = "{{ mlab_ns_query_url_ansible }}"
```
hellais: I think it's safe to keep this value hardcoded and not require configuring it from ansible as it's probably not going to be changing that often.

ansible/roles/ooni-backend/templates/update_bouncer.py.j2
```diff
@@ -0,0 +1,79 @@
...
+        f = open(path, 'w')
+        f.write(bouncer_config_contents)
+        f.close()
+    except IOError:
+        print "Couldn't write to bouncer config file."
+        exit(1)
+
+
+dst_bouncer_path = '{{ dst_bouncer_path_ansible }}'
+base_bouncer_path = '{{ base_bouncer_path_ansible }}'
```
hellais: I think it's better perhaps to convert these two options to something that is passed as arguments to the script as opposed to writing the script with some values inside of it.

It makes updating the script and checking for which version of it is on a given machine a bit more complex I think.

ansible/roles/ooni-backend/vars/main.yml
```diff
@@ -8,8 +8,20 @@ enable_daphn3_helper: yes
...
+dns_test_helper_address: "213.138.109.232:57004"
+http-return-json-headers_address: "http://93.95.227.200"
+ssl_address:
+tcp_echo_address: "213.138.109.232"
+traceroute_address: "213.138.109.232"
+web_connectivity_domain: "{{ letsencrypt_domain_list|selectattr('role', 'equalto', 'collector')|map(attribute='domain')|list }}"
+web_connectivity_hs: "{{ web_connectivity_hsdir_hostname.stdout }}"
+collector_hs: "{{ collector_hsdir_hostname.stdout }}"
+mlab_ns_query_url_ansible: "{{ http://mlab-ns.appspot.com/ooni?policy=all }}"
+dst_bouncer_path_ansible: "{{/data/bouncer/bouncer.yaml}}"
```
hellais: Are you sure this goes with `{{` around them?

ansible/roles/ooni-backend/tasks/cron.yml
```diff
@@ -0,0 +1,15 @@
+---
+- name: Template update bouncer script
+  template: >
+    src="update_bouncer.j2"
```
hellais: Based on the comments below this should probably not be a template, but rather just be a file that is copied.

ansible/roles/ooni-backend/templates/update_bouncer.py.j2
anadahz: I can change that but I don't really see the reason to do this since we will need to update the task whereas now we need to change only an inventory variable.

ansible/roles/ooni-backend/templates/update_bouncer.py.j2
anadahz: It is very important to have different variables in PATHs when you are deploying a fleet of backends, we need to be sure that this PATH is for sure the same that ooni-backend uses otherwise it's very hard to track down if there is something wrong.
If you hardcode this we need to hardcode most backend internal PATHs.

ansible/roles/ooni-backend/vars/main.yml
anadahz: Yes these are indeed variables!

ansible/roles/ooni-backend/tasks/cron.yml
anadahz: In our current stage/testing this would be very difficult to automate if we use hard-coded values.
Since we are going to be needing two cronjobs --a testing and a staging one that have different configuration and files.

anadahz:
This is ready to be merged fixes #55 .

ansible/roles/ooni-backend/templates/update_bouncer.py.j2
hellais: Yes, I understand, my suggestion was to instead of making the hardcoded value be generated by the ansible script to change the script to take arguments.

Something in the lines of:

```
import sys

# run this with python update-bouncer.yaml dst/bouncer/path.yaml base/bouncer/path.yaml
if __name__ == "__main__":
    dst_bouncer_path = "/data/bouncer/bouncer.yaml"
    base_bouncer_path = "/data/bouncer/bouncer-base.yaml"

    if len(sys.argv) == 3:
        dst_bouncer_path = sys.argv[1]
        base_bouncer_path = sys.argv[2]

    base_bouncer = yaml.safe_load(open(base_bouncer_path))
    while True:
        try:
            parts = read_parts_from_mlabns()
            break
        except:
            pass
    bouncer_config = assemble_bouncer_config(base_bouncer, parts)
    write_bouncer_config(dst_bouncer_path, bouncer_config)
```

ansible/roles/ooni-backend/vars/main.yml
```diff
@@ -3,13 +3,21 @@ enable_dns_helper: yes
...
 enable_tcp_echo_helper: yes
-enable_ssl_helper: no
 enable_daphn3_helper: yes
-enable_tls_bouncer: no
-enable_tls_collector: no
-enable_tls_web_connectivity: no
 
 dns_resolver_address: "8.8.8.8:53"
+dns_test_helper_address: "213.138.109.232:57004"
+http_return_json_headers_address: "http://93.95.227.200"
```
hellais: This test-helper appears to not be online anymore. I would either disable it or change it to something that is online (though I couldn't find any test-helper of this sort, except the mlab ones that change, running)

ansible/roles/ooni-backend/templates/update_bouncer.py.j2
```diff
@@ -0,0 +1,79 @@
...
+        f = open(path, 'w')
+        f.write(bouncer_config_contents)
+        f.close()
+    except IOError:
+        print "Couldn't write to bouncer config file."
+        exit(1)
+
+
+dst_bouncer_path = '{{ bind_bouncer_base_file }}'
+base_bouncer_path = '{{ bind_bouncer_file }}'
```
hellais: Resuming thread https://github.com/TheTorProject/ooni-sysadmin/pull/62#discussion_r68577390 here.

Also given the fact that the location of these files appears to always be the same (since it's a virtually mapped directory) I don't see in which case this would be something that would change.

hellais:
A part from the minor comments I think this looks good. 👍

ansible/roles/ooni-backend/vars/main.yml
anadahz: OK let's deploy some then, now that we have the alternate section can we also add multiple test-helpers in case some cannot be accessed or are blocked from the client side?

hellais: Currently the `test-helpers-alternate` only works for the web-connectivity test helper.

anadahz:
https://github.com/TheTorProject/ooni-sysadmin/pull/62/commits/cc3ffbf0d442cc1c88a202b0d95a9beab7dedcb8 addresses @hellais comments.

anadahz:
This fixes also #49 .

hellais:
Looks good too me!

PR #69, opened 2016-08-09
Ansible playbook obfs4 Tor bridge deployment

anadahz:
This role install the dependencies needed for an obfs4proxy bridge,
provisions the bridge and fetches the fingerprint and the generated
client torrc bridge line.

### DIFF
--- a/ansible/roles/tor_bridge/hosts
+++ b/ansible/roles/tor_bridge/hosts
@@ -1,0 +1,7 @@
+[bridges]
+37.218.240.79 tor_nickname=Joubini obfs4_listen_port=8080
+1.209.122.142 tor_nickname=Hapalochlaena obfs4_listen_port=5222
+199.119.112.170 tor_nickname=Bimaculoides obfs4_listen_port=5223
+
+[bridges:vars]
+tor_contactinfo=admin@openobservatory.org
--- a/README.md
+++ b/README.md
@@ -147,6 +147,20 @@ be used instead.
 ansible-playbook -i "hosts-inventory" ansible/install-letsencrypt.yml -v
 ```
 
+## install tor bridge (obfs4)
+
+This role install the dependencies needed for an obfs4proxy bridge, provisions
+the bridge and fetches the fingerprint and the corresponding generated client
+`torrc` configuration bridge line.
+This role should be running with the following tags:
+`install_tor`, `upgrades` and `install_obfs4`.
+
+### execute role
+
+```
+ansible-playbook -i hosts ansible/install-tor-bridge.yml --tags "install_tor,upgrades,install_obfs4"
+```
+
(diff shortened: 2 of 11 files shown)


hellais:
I am not sure that we want to include obfs4proxy bridge deployment ansible playbooks to the ooni-sysadmin repo. It seems like this could result in the scope of this repository just expanding too much.
Also have you looked at: https://github.com/david415/ansible-tor#example-tor-obfs4-bridge-playbook?

We could maybe include that as a submodule and just call that as a dependency.

anadahz:
@hellais https://github.com/david415/ansible-tor#example-tor-obfs4-bridge-playbook is non maintainable and according to its author the alternative https://github.com/nusenu/ansible-relayor should be user which I find too overloaded to just deploy some obfs4 bridges.

Given the fact that we do have some obfs4 bridges as part of our infrastructure I guess it makes perfectly sense for this piece of code to reside here.

ansible/roles/tor_bridge/hosts
```diff
@@ -0,0 +1,7 @@
+[bridges]
+37.218.240.79 tor_nickname=Joubini obfs4_listen_port=8080
+1.209.122.142 tor_nickname=Hapalochlaena obfs4_listen_port=5222
+199.119.112.170 tor_nickname=Bimaculoides obfs4_listen_port=5223
```
darkk: `37.218.240.79` and `199.119.112.170` seem to be still alive. Should they be included into `inventory` somehow? Are they _part of OONI infrastructure_?
`1.209.122.142` seems to be dead.

hellais:
I am going to close this as it's outdated.

ansible/roles/tor_bridge/hosts
hellais: `199.119.112.170` is now dead. `37.218.240.79` is not in the ooni eclipsis account so I think we can assume these as irrelevant.
@anadahz do you agree? If so I suggest we delete this branch.

anadahz: Please go ahead and delete this branch.
For reference two these bridges are still online (one on a different IP).

Does OONI has any obfs4 bridges running at the moment?

darkk: > Does OONI has any obfs4 bridges running at the moment

Nope, only tor nodes in some networks to act as test-helpers during response to some tor-blocking incident.

I currently see no point to run `obfs4` bridges as part of OONI infra as IP:Port based blocking can be debugged with Tor relay as well. And obfs4-specific protocol-based blocking does not need a bridge that was online "long enough" to get blocked.

darkk:
I'm unsure why it was reopened.
Seems, the consensus is that this PR is not currently relevant.

PR #72, opened 2016-10-05
Add docker-py version requirement, set state,restart policy

anadahz:
Ansible docker_py module has deprecated the build state and bumped the
docker-py dependency to >= 1.7.0. Adding latest to the version doesn't
work as expected as the docker_py is not correctly validating the
version number.

The module option 'build' in state is deprecated and will now behave the
same as the 'present' option.

Set the docker container policy to always, to restart stopped containers


hellais:
LGTM

PR #75, opened 2016-11-02
Update ooniprobe role to install ooniprobe v2

anadahz:
* Install ooniprobe v2 via pip and run ooniprobe-agent as a service with
systemd.
* Improve installation of tor GPG key
* Update documentation


darkk:
@hellais @anadahz is this code still useful?

`ooniprobe` role is deployed to `local:custom` set of nodes that are not found in the `inventory`.

anadahz:
> @hellais @anadahz is this code still useful?

This role has been used to deploy ooniprobe in Debian like systems.
It's not part of the infrastructure though used to be useful to deploy ooniprobe via ansible.

> ooniprobe role is deployed to local:custom set of nodes that are not found in the inventory.

The inventory groups have changed significantly since then.
The `custom` group name was used as a set of custom nodes defind by the user.

hellais:
@anadahz can we drop this PR? I would drop it since it's not really working anymore properly.

anadahz:
Indeed let's drop this PR.

PR #77, opened 2016-11-20
Add ansible PostgreSQL installation role

anadahz:


ansible/install-postgresql.yml
```diff
@@ -0,0 +1,14 @@
+---
+- hosts: localhost
+  gather_facts: no
+  vars:
+    instance_name: ooni-measurements-db
+    disk_size: 200
+    ram_size: 4096
+    cpu_cores: 2
+  roles:
+    - { role: eclipsis, when: api_token is defined }
```
hellais: What role are you referencing here?

anadahz: This one: https://github.com/TheTorProject/ooni-sysadmin/tree/eclipsis/ansible-roles/ansible/roles/eclipsis which is not really that well tested.

hellais: Should probably also merge that since this doesn't make sense without it.

anadahz: It makes sense though eclipsis role is not a heavy dependency to run this role; one can still use to setup the postgresql role.

Issue #74, opened 2016-10-26
Make the canonical collector and test helpers listen on port 443


hellais:
It seems like it's fairly common for some networks (such as mobile ones) to block access to anything that is not traffic to common ports.

Currently the canonical collector binds on port 4441 and the canonical web connectivity test helpers binds on 4442.

It would be great if we could setup another machine just to host the collector and web connectivity test helper that has a dedicated IP for each of those services.

The added benefit of this would be that of being able to divert clients to another working collector from the bouncer (that would then reside on it's own machine) in case we encounter difficulties with it.

anadahz:
In order for the deployment to take place we 'll be needing a bouncer base file pointing to all collectors, and test helpers instances, such as the example below:

```
collector:
  httpo://s2zznrn5jxfgkaom.onion:
    collector-alternate:
    - { address: 'https://a.collector.stage.ooni.io:4441', type: 'https'}
    - { address: 'http://a.collector.stage.ooni.io:8081', type: 'http'}
    - { address: 'https://{{ collector_domain }}:{{ tls_collector_port }}', type: 'https'}
    - { address: 'http://{{ collector_domain }}:{{ tcp_collector_port }}', type: 'http'}
    test-helper:
      {dns: '213.138.109.232:57004',
       http-return-json-headers: 'http://93.95.227.200',
       tcp-echo: '213.138.109.232',
       web-connectivity: 'ebckmavdjlxhnw27.onion',
       traceroute: '213.138.109.232'}
    test-helper-alternate:
      web-connectivity:
      - {address: 'ebckmavdjlxhnw27.onion', type: 'onion'}
      - {address: 'https://a.web-connectivity.th.stage.ooni.io:4442', type: 'https'}
      - {address: 'http://a.web-connectivity.th.stage.ooni.io:8082', type: 'http'}
      - {address: 'https://{{ web_connectivity_domain }}:{{ tls_web_connectivity_port }}', type: 'https'}
      - {address: 'http://{{ web_connectivity_domain }}:{{ tcp_web_connectivity_port }}', type: 'http'}
```

hellais:
I see that you are specifying in here also staging collectors and test helpers. We should exclude these from the production collector and just include the ones that are production.

I would imagine the bouncer file for the production deployment would become something like this:

**current bouncer.yaml**
```json
collector:
  httpo://ihiderha53f36lsd.onion:
    collector-alternate:
[...]

PR #79, opened 2016-12-16
Add oonibackend deployment on separate containers

anadahz:
* Add task to install docker-engine
* Add tasks and change default role behavior to enable installation of
  oonibackend bouncers, collectors and test-helpers in separate instances
  that can be deployed in different hosts and containers


anadahz:
re: #74

ansible/deploy-oonibackend-docker.yml
```diff
@@ -24,6 +77,7 @@
     enable_tcp_web_connectivity: yes
     enable_tcp_bouncer: yes
     enable_tcp_collector: yes
+    oonibackend_bundle: yes
```
hellais: I guess if I have to nitpick it's a bit confusing that setting this variable to "no" actually doesn't have the expected effect of disabling it, since you just check for the variable being defined.

ansible/roles/letsencrypt/tasks/letsencrypt_renew.yml
```diff
@@ -4,5 +4,5 @@
     name: "Renew ooni-backend letsencrypt certificates"
     user: "root"
     cron_file: "letsencrypt_renew_certs-{{ app_env }}"
-    special_time: "monthly"
+    special_time: "weekly"
```
hellais: After how much time since a let's encrypt cert is issues can you renew it? I suspect this being monthly was the reason why it was not working as expected in the past.
Maybe we can even go as far as doing it daily.

Also, what happens when this role is run?

Is the oonibackend taken offline while it is running?

hellais:
Added a couple of comments, when they are addressed this LGTM

ansible/roles/letsencrypt/tasks/letsencrypt_renew.yml
anadahz: This cronjob will check if renewal is necessary and will try to renew the required certificates if it's needed (i.e certificate's expiration date is approaching).
During the renewal checks there is no service interruption.

When the certificates are due for renewal, certbot (letsencrypt's component) will start the renewal process by bringing down the appropriate container, renew the certificates, copy the updated certificates in oonibackend's certificate storage and then start the appropriate container for oonibackend to read the updated certificates.

Currently the deployed oonibackend service is set to check for certificate renewals monthly, we can decrease the renewal time period to daily, though setting the cronjob to run hourly, weekly or monthly has nothing to do with #80 (if this is the bug mentioned by not working as expected).

The oonibackend services need to be taken offline in order for oonibackend to read the newly generated certificates (see: https://github.com/TheTorProject/ooni-backend/issues/90).

ansible/roles/ooni-backend/tasks/populate_bouncer_base.yml
```diff
@@ -15,13 +15,32 @@
...
+  register: bouncer_config_file
 
-- name: Fetch backend related config files and HS keys
+- name: Template "{{ bind_bouncer_file }}" with "{{ bouncer_base_template }}"
+  template: >
+    src="{{ bouncer_base_template }}"
+    dest="{{ bind_bouncer_file }}"
+    backup="yes"
+    unsafe_writes="true"
+  when: ( oonibackend_bundle is undefined or oonibackend_bundle == False )
```
hellais: I believe you need to cast this to bool and check for `False` with `False` written in lowercase:

like this: `oonibackend_bundle|bool == false`.

You can also avoid having to do the undefined check by setting a default for it like so:

`oonibackend_bundle|default(false)|bool == false`

anadahz: @hellais is there any case where the conditional for `oonibackend_bundle` variable in https://github.com/TheTorProject/ooni-sysadmin/pull/79#discussion-diff-94381954R33 fails to evaluate when `oonibackend_bundle` is unset or set to: Yes/yes, No/no, True/true, False/false ?
If not I will prefer to use the same conditional syntax as in most (if not all) ansible roles/playbooks in this repository.

hellais: > is there any case where the conditional for oonibackend_bundle variable in #79 (diff) fails to evaluate when oonibackend_bundle is unset or set to: Yes/yes, No/no, True/true, False/false ?

Yes, in old version of jinja2, where it was incorrect to say "False" instead of "false", it seems like recent versions don't behave that way anymore, but it's still incorrect to use "False" instead of "false": http://jinja.pocoo.org/docs/dev/templates/#expressions.

> If not I will prefer to use the same conditional syntax as in most (if not all) ansible roles/playbooks in this repository.

I don't see this pattern used anywhere else in the repo actually (the check for true false is only used here), while the is undefined check is used in only 7 places:
```json
$ grep -rn 'is undefined' ansible | wc -l
       7
```

On the other hand the casting to a default value is used in many more places:
```
$ grep -rn 'default(' ansible | wc -l
      24
```

I would change the 7 lines above to use this better pattern.

anadahz: > On the other hand the casting to a default value is used in many more places:

> $ grep -rn 'default(' ansible | wc -l
>      24

In your [comment](https://github.com/TheTorProject/ooni-sysadmin/pull/79#discussion_r94381954) you mentioned that I should instead cast to bool.
```
 grep -rn 'bool' ansible | wc -l
 0
 ```
 Also the default filter is only used in templates:
 ```
 grep --exclude-dir=templates -rn 'default(' ansible | wc -l
 0
 ```

 I can change all the conditionals of this repository to your preferred syntax pattern but I believe this is not related and should not be addressed in this PR.

hellais: Sure, I guess my point was that there are usages of the default filter. Just use in this one the better syntax of casting to bool that is more robust and compact and you can address the other ones in another pull request.

PR #81, opened 2016-12-30
Add ooni-measurements deployment ansible role

anadahz:


hellais:
LGTM

hellais:
(BTW LGTM mean Looks Good To Me or Looks Good To Merge, so when me or anybody else writes that it means that you can proceed with merging the branch without any further need for feedback).

PR #82, opened 2017-01-11
Do not try to fetch files unrelated to the instance

anadahz:


anadahz:
Going to self merge this.

PR #85, opened 2017-01-28
Fix conditionals, docker image rebuild and deploy on success

anadahz:
* The oonibackend docker images should always rebuild and deployed upon
  successful build.
* Fix playbook conditionals


anadahz:
Going to process and merge this.

PR #105, opened 2017-04-18
Airflow pipeline in da docker (part 1)

darkk:
It's WIP, but you may already have some comments.

BTW, does Travis check something useful for ooni-sysadmin?

Stuff I'll do before mering (some of that will be done while working on tiling puzzles):
- [x] pin exact `openobservatory/shovel` version
- [x] put `/srv/airflow-dags` under version control
- [x] understand the reason for `airflow service -- {worker,scheduler}` tasks constantly reporting `changed`
- [x] link to [ansible vault BCP](http://docs.ansible.com/ansible/playbooks_best_practices.html#best-practices-for-variables-and-vaults)
- [x] s/$priv/$private/; s/$pub/$public/ (for grepability);
- [x] drop `bootstap python` pre_task
- [x] s/accoints/accounts/; s/furhter/further/ ; s/exemples/examples/
- [x] `restart_policy: unless-stopped`
- [x] `sleep 60` in async
- [x] name for ansible tasks
- [x] s/cp -a/M(copy) remote_src=True/


ansible/host_vars/datacollector.infra.ooni.io/vault
```diff
@@ -0,0 +1,26 @@
...
+63363730386332356431653433313838646138623634656538656633623830356464386330656634
+38306434643365346639326430383432313238613338663130623138646130356364346432333163
+32396236366663616134383863366135666538376562313738356362366662376135336635363536
+37313663333065333666626332393233663231306561663338373130616166386533313761353063
+63323334386165356535626239393663383136313861333534393231333465643730613831366561
+39346663353864363436343734363230656334326430363034323936633231316438666137353263
+62343765343163303536336335383936343462616232353639383866396432373363666236356136
+63663463373632306233356366313838613233656263343030363131323534303935303337663363
+35633666643130336530663338653331346634636234386339346366306366373733333131643837
+3965
```
hellais: 🎉

ansible/group_vars/all/vars.yml
```diff
@@ -0,0 +1,49 @@
...
+  andz:
+    login: andz
+    id: 10001
+    comment: Vasilis Ververis
+    keys: ["ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIE4xZd3VJ5+EHGSd4mZIdXBqUQwtABsEhyrGHYyFyddZ andz"]
+  art:
+    login: art
+    id: 10002
+    comment: Arturo Filasto
+    keys: ["ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEAu4HGXj2rXvZhVwts/F/yeF+G+OFAuhRhKNtAD+F8xu8z2flsNnW13u9Aspe53uMafSYYrujLYu6pm8UE/MZ5XiGaRRivQVhn/qGxv8HQOaYWihn3npTyX9p10W8xDm45P1n5EGV8NSgcE5UZ3MDDwT8uh9oIsq3VFAUea4yn9V/l9ZImGamfEiQlXQ3Jj9KoKH9EgTvh0FQmnJtMldm32T5ZAZtJ/TGWdwr3QBfyh8nZoCKHqO8c9qBB2nJRR1t0O73/Cy+tmWdWsFpSeVps2wg8THfmdm+TWTno6iD+uTmh8pALmuyuwcVzU++xaiE65GmocJlFF/tSS1oOeDI8jw== art@torproject.org"]
```
hellais: I would prefer this key be used instead for eclipsis hosts:
```
ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQC9her6lZIfP/ulXleu00u3ukdcVEDslkfXzi65JFRXRkKYsdMw7zLvrkvFryqDGsRQwpZo0rlk5o26QNSUR/g2Kf3wEHxxmMcN4+g+f0/fJDKLHSFdL5Z5ACyp5YvjA88vG753qVn7lBINuvCb4wk1sUOWM1WVG5WiszLiJCddYUrgCZWzD3FHyApJ/GGH2A5vTTN8mcQapQaJ9/sgxDw5WT+yMtZ8TW4462x+TaWWGE0WnAd/l6HOFaS14mql0K+MbtRp9EKN2O6+sA0OsjknYpt21ovtEWfBiojjh+4Iz6bTk+M5+fWHwQaxqEcaok22l35GudlmaoTB0OesB/1z art@eclipsis
```

ansible/inventory
```diff
@@ -1,5 +1,37 @@
...
+#c.web-connectivity.th.ooni.io
+b.collector.ooni.io
+
+[hkg]
+datacollector.infra.ooni.io
+# irc-bouncer.service.ooni.io # keep commented for a while, broken UIDs there
+
+[ams]
+c.collector.ooni.io
+#measurements.ooni.io
```
hellais: We also have a.collector.ooni.io

darkk: I consider bouncer being canonical name for that box:
```
a.collector.ooni.io has address 62.217.125.6
bouncer.infra.ooni.nu has address 62.217.125.6
```

darkk:
> - [x] understand the reason for airflow service -- {worker,scheduler} tasks constantly reporting changed

Ansible 2.2.2 does not recurse into data structures passed to modules while checking for `omit`, only first-level arguments can be [omitted](http://docs.ansible.com/ansible/playbooks_filters.html#omitting-parameters) with `omit` value, see ansible/ansible#14130 for details.

```diff
diff --git ansible/roles/airflow/tasks/af-container.yml ansible/roles/airflow/tasks/af-container.yml
index f9f0ade..94286ac 100644
--- ansible/roles/airflow/tasks/af-container.yml
+++ ansible/roles/airflow/tasks/af-container.yml
@@ -11,9 +11,14 @@
     hostname: "{{ dockname }}"
-    networks: [{name: af, ipv4_address: "{{ ipv4_address | default(omit) }}" }]
+    networks: [ "{{ {'name': 'af'} | combine(
+      ipv4_address | default(False) | ternary({'ipv4_address': ipv4_address}, {})
+    ) }}" ]
     purge_networks: true
```

Previous version was producing following container configuration according to `docker inspect`:
```json
"Networks": {
  "af": {
    "IPAMConfig": {
      "IPv4Address": "__omit_place_holder__59f22fb877f2a710e45de3fa900db77646704530"
    },
```
Seems, docker 1.12 was happy with IPv4Address that is neither IPv4 nor valid DNS name, it seems to be [fixed in 1.13+](https://github.com/moby/moby/commit/c7045eb93523df76feb9f3e00540e69471a1855e).

ansible/inventory
```diff
@@ -1,5 +1,87 @@
...
+[ath] # Athens. Greek Research and Technology Network S.A
+bouncer.infra.ooni.nu # NOT-GH, a.collector.ooni.io a.web-connectivity.th.ooni.io a.collector.ooni.io bouncer.ooni.io
+puppet-master-1.infra.ooni.io # NOT-GH, java1.interoperability.gr
+
+[ber] # Humboldt-Universitaet zu Berlin
+chameleon.infra.ooni.io # NOT-GH
+polaris.infra.ooni.io # NOT-GH, jumphost=chameleon, tune ~/.ssh/config
+
+[quadix] # QuadIX (QUADI-8), US, Pensilvania
+pipeline.infra.ooni.nu # NOT-GH
```
hellais: This is long dead.

ansible/inventory
```diff
@@ -1,5 +1,91 @@
...
+hkg
+ams
+ath
+ber
+quadix
+tpo
+bigv
+
+[wdc]
+mail.ooni.io # NOT-GH, Art-No-ACCESS - no idea what this is
```
anadahz: Not used.

ansible/inventory
```diff
@@ -1,5 +1,91 @@
...
+mail.test.ooni.io # NOT-GH, Art-No-ACCESS, mail-test.ooni.io - no idea what this is
```
anadahz: Not used.

ansible/inventory
```diff
@@ -1,5 +1,91 @@
...
+ooni-fdroid-build.infra.ooni.io # Art-NO-ACCESS
+a.echo.th.ooni.io # OK (u: root)
+separate.b.collector.test.ooni.io # NOT-GH, Art-NO-ACCESS, OFFLINE?
```
anadahz: Not used.

anadahz: Not used.

ansible/inventory
```diff
@@ -1,5 +1,91 @@
...
+separate.bouncer.test.ooni.io # NOT-GH, Art-NO-ACCESS, OFFLINE?
+separate.a.web-connectivity.th.test.ooni.io # NOT-GH, Art-NO-ACCESS, OFFLINE?
```
anadahz: Not used.

ansible/inventory
```diff
@@ -1,5 +1,91 @@
...
+ooni-proxy-dc.infra.ooni.io # OK - MAYBE DISCONTINUE?
+ooni-backend-testing.infra.ooni.io # OK - Temporary machine
+oonibackend-stress-test.infra.ooni.io # OK - MAYBE DISCONTINUE?
+apk-build-test.infra.ooni.ioa # Art-NO-ACCESS - no idea what this is
+
+[hkg]
+datacollector.infra.ooni.io
+# irc-bouncer.service.ooni.io # OK, keep commented for a while, broken UIDs there
+vpn-gate-runner.infra.ooni.io # OK
+seedling.infra.ooni.io # Art-NO-ACCESS - I think this is the machine used to seed lepidopter torrent
```
anadahz: This has been used to seed the lepipodpter torrent, as we decided not offer any torrent we are going to deprecate this host.

ansible/inventory
```diff
@@ -1,5 +1,91 @@
...
+c.echo.th.ooni.io # Art-NO-ACCESS, CRITICAL
+reports.ooni.nu # vps770.greenhost.nl NOT-GH - DISCONTINUED
+
+[ath] # Athens. Greek Research and Technology Network S.A
+bouncer.infra.ooni.nu # NOT-GH, OK, a.collector.ooni.io a.web-connectivity.th.ooni.io a.collector.ooni.io bouncer.ooni.io
+puppet-master-1.infra.ooni.io # NOT-GH, Art-NO-ACCESS, java1.interoperability.gr - Is this still even alive?
+
+[ber] # Humboldt-Universitaet zu Berlin
+chameleon.infra.ooni.io # NOT-GH, OK
+polaris.infra.ooni.io # NOT-GH, OK, jumphost=chameleon, tune ~/.ssh/config
```
anadahz: Discontinued!

ansible/inventory
```diff
@@ -1,5 +1,91 @@
...
+
+[tpo] # http://www.euserv.de
+deb.ooni.nu # NOT-GH, OK - This can probably be discontinued
+staticiforme.torproject.org # - art, weasilis have access. (bounce via perdulce.tpo). This requires tpo LDAP account and various other fluff.
+
+[bigv] # Bytemark @ London
+b.echo.th.ooni.io # NOT-GH, down, mirror1.reports.ooni.nu ooni-1.default.orgtech.uk0.bigv.io, OK (u: $login)
+
+[scaleway]
+rpi-build # andz: is it OK to publish the IP of this host (last octet .250)?
```
anadahz: This should be removed for the inventory, non OONI infrastructure.

anadahz:
Thanks for working on this @darkk let me know when you are ready with it so that I can review and test the ansible playbook/roles.

darkk:
I suggest for everyone interested to review this branch so we can merge it to `master` and base further branches on top of master. I'd rather move unchecked items in separate "unit of review" instead of planting a forest of branches.

ansible/host_vars/localhost/vars.yml
```diff
@@ -0,0 +1,5 @@
+---
+gh_token: '{{ vault_gh_token }}'
+namecheap_user: '{{ vault_namecheap_user }}'
+namecheap_token: '{{ vault_namecheap_token }}'
+namecheap_jumphosts: '{{ vault_namecheap_jumphosts }}'
```
hellais: Note to self. Adjust my roles using the `vault_` convention when referencing vaulted variable (maybe also add it to readme).

ansible/inventory
```diff
@@ -1,5 +1,90 @@
...
+vps770.greenhost.nl # DELETE discontinued, old-collector
+#211.149.218.16 old-probe
+87.107.121.133 # DELETE discontinued, old-probe
+mail.ooni.io # DELETE
+mail.test.ooni.io # DELETE
+separate.b.collector.test.ooni.io # DELETE
+separate.bouncer.test.ooni.io # DELETE
+separate.a.web-connectivity.th.test.ooni.io # DELETE
+seedling.infra.ooni.io # DELETE
+reports.ooni.nu # vps770.greenhost.nl NOT-GH - DELETE
```
hellais: If we agree we can delete these from here and drop the A records where we can.

ansible/roles/airflow/files/airflow-dags/canning.py
```diff
@@ -0,0 +1,71 @@
...
+dag.set_dependency('autoclaving', 'meta_pg')
+
+dag.set_dependency('reports_raw_s3_ls', 'reports_raw_cleanup')
+dag.set_dependency('canning', 'reports_raw_cleanup')
+
+dag.set_dependency('autoclaving', 'sanitised_check')
+
+dag.set_dependency('autoclaving', 'sanitised_cleanup')
+dag.set_dependency('sanitised_s3_ls', 'sanitised_cleanup')
+dag.set_dependency('sanitised_check', 'sanitised_cleanup')
```
hellais: Shouldn't this file live inside of ooni-pipeline?

ansible/roles/airflow/files/docker-trampoline
```diff
@@ -0,0 +1,121 @@
...
+        exit 1
+    fi
+done
+) || exit 1
+
+docksafe="--rm --interactive --read-only"
+docknet="--network none"
+
+priv=/data/ooni/private
+pub=/data/ooni/public
```
hellais: Maybe these need to be parametrised or if not at least mention in the comment in the header that this is the expected filesystem layout.

ansible/inventory
```diff
@@ -1,5 +1,90 @@
...
+b.collector.ooni.io
+
+# The machines below maybe can be discontinued
+ooni-proxy-dc.infra.ooni.io
+ooni-backend-testing.infra.ooni.io
+oonibackend-stress-test.infra.ooni.io
+
+tipitpitap.infra.ooni.io # Art-NO-ACCESS - no idea what this is
+ooni-fdroid-build.infra.ooni.io # Art-NO-ACCESS
+apk-build-test.infra.ooni.io # Art-NO-ACCESS - no idea what this is
```
hellais: This are still unclaimed. We maybe want to move some of them to the delete section.

ansible/host_vars/localhost/vars.yml
darkk: WRT Readme — link to ansible BCP may be added.

ansible/roles/airflow/files/airflow-dags/canning.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
```
hellais: I wonder if actually all the files listed in `airflow-dags` should actually not just live inside of ooni-pipeline. If not what is the reason for doing so?

ansible/inventory
darkk: That's exactly what I was going to do. But I'd rather keep them in history.

hellais:
I left some very minimal comments, but overall this PR is good to be merged for me. I have also been testing some of the bits of it as part of the monitoring deployment stuff, so for me this can be merged.

ansible/roles/airflow/files/airflow-dags/canning.py
darkk: Maybe it would be more pure, but maintaining it as a "glue" code in ooni-sysadmin makes update easier: you don't have to pinpoint exact commit to fetch, you don't have to deal with github inability to export set of raw files for specific commit, etc.
So I decided to place "actions" to ooni-pipeline and all the glue with paths and dependencies to ooni-sysadmin to avoid getting to meta.

ansible/roles/airflow/files/airflow-dags/canning.sh
darkk: It's same deployment simplicity reason. Everything that depends on exact paths goes to sysadmin repo. These launcher bash scripts are essentially hard-coded configuration variables passed as CLI arguments

darkk: It's same deployment simplicity reason. Everything that depends on exact paths goes to sysadmin repo. These launcher bash scripts are essentially hard-coded configuration variables passed as CLI arguments

ansible/dom0-bootstrap.yml
```diff
@@ -0,0 +1,18 @@
+---
+- include: ansible-version.yml
+
+- hosts: gh # TODO: it eventually should become `all`
+  become: false
+  remote_user: root
+  gather_facts: false
+  pre_tasks:
+    - name: bootstap python
+      raw: if [ ! -x /usr/bin/python ]; then apt-get update && apt-get -y install python-simplejson python-apt; fi
```
anadahz: There is no need to "manually" install the package `python-apt`, ansible will [install]((https://github.com/ansible/ansible/pull/4617)) if/when it's needed.

ansible/ansible-version.yml
```diff
@@ -0,0 +1,14 @@
+---
+- hosts: all
```
anadahz: I think that this will not test the localhost which is the only that this roles wants to test?

ansible/inventory
anadahz: Yep, go ahead and remove the A records but please keep them in history.

ansible/roles/adm/tasks/main.yml
```diff
@@ -0,0 +1,41 @@
+---
+# This role is part of bootstrap process that creates sysadmin accoints so
```
anadahz: minor typo: s/accoints/accounts

ansible/roles/adm/tasks/main.yml
```diff
@@ -0,0 +1,41 @@
...
+# furhter setup can be done with logged `sudo`.  ooni-sysadmin repo is single
```
anadahz: minor typo: s/furhter/further

ansible/roles/adm/tasks/main.yml
```diff
@@ -0,0 +1,41 @@
...
+    label: "{{ item.login }}"
+
+- name: create .ssh/authorized_keys
+  template: src=authorized_keys dest=/home/{{item.login}}/.ssh/authorized_keys owner={{item.login}} group={{item.login}} mode=0400
+  with_items: "{{ adm_passwd }}"
+  loop_control:
+    label: "{{ item.login }}"
+
+- name: sudoers.d/adm
+  template: src=sudoers dest=/etc/sudoers.d/adm owner=root group=root mode=0440 validate='visudo -cf %s'
```
anadahz: Wouldn't perhaps make more sense to add a password sydo group to the users that will have password less sudo access?
We could then (perhaps) use the `lineinfile` module and have one less template file.

ansible/roles/airflow/tasks/af-container.yml
```diff
@@ -0,0 +1,36 @@
...
+      EXECUTOR: Celery
+      FERNET_KEY: "{{ airflow_fernet_key }}"
+      REDIS_HOST: af-redis
+      REDIS_PASSWORD: "{{ airflow_redis_password }}"
+      POSTGRES_HOST: af-psql
+      POSTGRES_USER: airflow
+      POSTGRES_PASSWORD: "{{ airflow_postgres_password }}"
+    command: "{{ command }}"
+    user: "{{ passwd[login].id }}:{{ passwd[login].id }}"
+    restart_policy: on-failure
```
anadahz: This should be `restart_policy: always` if we want that the docker container restarts on boot

ansible/roles/airflow/tasks/main.yml
```diff
@@ -0,0 +1,155 @@
...
+    name: af-redis
+    hostname: af-redis
+    networks: [{name: af}]
+    purge_networks: true # drop `bridge` network
+    volumes:
+      - /srv/etc/af-redis:/usr/local/etc:ro
+      - /srv/af-redis:/data:rw
+    command: /usr/local/etc/redis.conf
+    user: "{{ passwd.afredis.id}}:{{ passwd.afredis.id}}"
+    restart_policy: on-failure
```
anadahz: This should be restart_policy: always if we want that the docker container restarts on boot

ansible/roles/airflow/tasks/main.yml
```diff
@@ -0,0 +1,155 @@
...
+    purge_networks: true # drop `bridge` network
+    volumes:
+      - /srv/tmp/af-psql:/run/postgresql:rw
+      - /srv/af-psql:/srv/af-psql:rw
+    env:
+      PGDATA: /srv/af-psql
+      POSTGRES_USER: airflow
+      POSTGRES_PASSWORD: "{{ airflow_postgres_password }}"
+    user: "{{ passwd.afpsql.id }}:{{ passwd.afpsql.id }}"
+    restart_policy: on-failure
```
anadahz: This should be restart_policy: always if we want that the docker container restarts on boot

ansible/roles/docker_py/tasks/main.yml
```diff
@@ -1,11 +1,10 @@
 ---
 
-- name: installs apt-https-transport and required kernel packages
+- name: installs apt-https-transport
   apt: >
     name={{ item }}
     state=present
   with_items:
-    - "linux-image-{{ ansible_kernel }}"
```
anadahz: This package was required in GH servers in order to install the docker engine.

ansible/roles/docker_py/tasks/main.yml
```diff
@@ -20,16 +19,21 @@
...
+#    https://github.com/ansible/ansible-modules-core/issues/5515
+#    https://github.com/ansible/ansible/issues/20776
+# 2. 2.0 is probably unsupported due to renaming of `docker.Client` to `docker.APIClient`
+#    https://docker-py.readthedocs.io/en/stable/change-log.html
 - name: Install docker-py via pip in virtualenv
   pip: >
     name="docker-py"
-    version="1.7.0"
-    virtualenv="{{ ansible_env.HOME }}/venv"
+    version="1.10.6"
```
anadahz: I remembered I had issues with any version greater that `1.7.0` perhaps this is fixed now.
See also: https://github.com/TheTorProject/ooni-sysadmin/commit/139cfab43a931db1aa39a46c7cf7cb7abedc8439

docker/airflow/entrypoint.sh
```diff
@@ -0,0 +1,107 @@
...
+: ${POSTGRES_DB:="airflow"}
+
+: ${FERNET_KEY:=$(python -c "from cryptography.fernet import Fernet; FERNET_KEY = Fernet.generate_key().decode(); print(FERNET_KEY)")}
+
+if [ -n "$FERNET_KEY" ] && ! python -c 'import cryptography.fernet, sys; cryptography.fernet.Fernet(sys.argv[1])' "$FERNET_KEY"; then
+    echo "$(date) - malformed FERNET_KEY disables encryption"
+    exit 1
+fi
+
+# Load DAGs exemples (default: Yes)
```
anadahz: minor typo: s/exemples/examples

ext/hosts
```diff
@@ -0,0 +1,56 @@
...
+            dns.setdefault(el.get('Address'), set()).add(el.get('Name') + '.' + domain)
+
+droplets = json.load(open('gh-droplets.json'))
+for vm in droplets['droplets']:
+    for net in vm['networks']['v4']: # no IPv6 support there yet
+        gh[net['ip_address']] = vm['name']
+
+all_ips = dns.viewkeys() | gh.viewkeys()
+
+nmap = check_output(['sudo', 'nmap', '-v', '-sn', '-oX', '-'] + list(all_ips))
```
anadahz: Is  the nmap package installed already?

ext/ooni.io.xml
```diff
@@ -0,0 +1,77 @@
...
+      <host HostId="112976293" Name="ooni-proxy-dc.infra" Type="A" Address="199.119.112.184" MXPref="10" TTL="1799" AssociatedAppTitle="" FriendlyName="" IsActive="true" IsDDNSEnabled="false" />
+      <host HostId="112976294" Name="ooni-backend-testing.infra" Type="A" Address="199.119.112.211" MXPref="10" TTL="1799" AssociatedAppTitle="" FriendlyName="" IsActive="true" IsDDNSEnabled="false" />
+      <host HostId="112976295" Name="oonibackend-stress-test.infra" Type="A" Address="199.119.112.233" MXPref="10" TTL="1799" AssociatedAppTitle="" FriendlyName="" IsActive="true" IsDDNSEnabled="false" />
+      <host HostId="112991539" Name="polaris.infra" Type="A" Address="141.20.103.26" MXPref="10" TTL="1799" AssociatedAppTitle="" FriendlyName="" IsActive="true" IsDDNSEnabled="false" />
+      <host HostId="85965209" Name="www" Type="URL" Address="https://ooni.torproject.org/" MXPref="10" TTL="1799" AssociatedAppTitle="" FriendlyName="" IsActive="true" IsDDNSEnabled="false" />
+    </DomainDNSGetHostsResult>
```
anadahz: How this file is being generated?
Is the `HostId` a "random" number?

ansible/ext-inventory.yml
```diff
@@ -0,0 +1,65 @@
...
+      command: git rev-parse --show-toplevel
+      register: extroot
+      changed_when: false
+
+    - set_fact:
+        extroot: '{{ extroot.stdout }}/ext'
+        socks5: '{{ 32767 | random(1024) }}'
+        jumphost: '{{ namecheap_jumphosts | random }}'
+    - name: ssh jumphost
+      command: ssh -D {{ socks5 }} {{ jumphost }} sleep 60
```
anadahz: I usually add the `-Nq` options to ssh jumphost commands as there is no need to execute a remote command to the jumpohost or be that verbose.

ansible/ext-inventory.yml
```diff
@@ -0,0 +1,65 @@
...
+  tasks:
+    - name: get git-root
+      command: git rev-parse --show-toplevel
+      register: extroot
+      changed_when: false
+
+    - set_fact:
+        extroot: '{{ extroot.stdout }}/ext'
+        socks5: '{{ 32767 | random(1024) }}'
+        jumphost: '{{ namecheap_jumphosts | random }}'
```
anadahz: I would prefer if there was one host specifically configured to be a jumphost, perhaps we should add such a role to the playbook.

ansible/play
```diff
@@ -0,0 +1,21 @@
...
+    fi
+  elif [ "$uname" = Darwin ]; then
+    name=`echo "play" "$@" | sed s/[^-+=a-zA-Z0-9_]/_/g`.`date +%Y.%m.%d.%H.%M.%S`.$$.log
+    exec script -a "$name" ansible-playbook --vault-password-file ./password-pipe --inventory-file ./inventory "$@"
+  else
+    echo "$0: don't know how to launch \`script\` at $uname." 1>&2
+    # fall-through
+  fi
+fi
+exec ansible-playbook --vault-password-file ./password-pipe --inventory-file ./inventory "$@"
```
anadahz: Shouldn't `./inventory` -> `./ansible/inventory` and `./password-pipe` -> `./ansible/password-pipe` in this file?

ansible/ext-inventory.yml
anadahz: Does the sleep command needs to be executed in the jumphost host?

anadahz:
@darkk How one should be running this playbook to test it?

ansible/roles/airflow/files/airflow-dags/canning.py
```diff
@@ -0,0 +1,71 @@
...
+        elif retcode == 13:
+            return False
+        else:
+            raise AirflowException('Unexpected `is-reports-raw-ready` exit code: {:d}'.format(retcode))
+
+dag = DAG(
+    dag_id='hist_canning',
+    schedule_interval=timedelta(days=1),
+    start_date=datetime(2012, 12, 5),
+    end_date=datetime(2017, 4, 22), # NB: end_date is included
```
anadahz: What does the end_date in this context means?

ansible/roles/airflow/tasks/af-container.yml
```diff
@@ -0,0 +1,36 @@
+---
+- file: path=/srv/{{ dockname }} state=directory mode=0750 owner={{ login }} group={{ login }}
+- stat: path=/srv/{{ dockname }}/airflow.cfg
+  register: afcfg
+- command: cp -a /srv/etc/af-share-{{ airflow_tag }}/airflow.cfg /srv/{{ dockname }}/airflow.cfg
```
anadahz: Is there any specific that the cp command is being used instead of the copy module?

anadahz:
Perhaps it makes more sense to add all the POSTGRES_* variables of the airflow role to `var/main.yml` so that they can be handled by ansible and more easily managed. The entrypoint.sh can be also templated by ansible so that the correct variables are being set for all tasks and scripts.

ansible/roles/airflow/tasks/main.yml
```diff
@@ -0,0 +1,155 @@
...
+- group: name={{ item.login }} gid={{ item.id }} state=present
+  with_items: "{{ svc_passwd }}"
+- user:
+    name: "{{ item.login }}"
+    group: "{{ item.login }}"
+    uid: "{{ item.id }}"
+    comment: "{{ item.comment }}"
+    shell: /usr/sbin/nologin
+    createhome: no
+    home: /nonexistent
```
anadahz: I guess the `home` option is redundant here.

ansible/roles/airflow/tasks/main.yml
```diff
@@ -0,0 +1,155 @@
...
+
+- name: docker network for airflow
+  docker_network:
+    name: af
+    driver_options:
+      com.docker.network.bridge.name: braf
+    ipam_options:
+      subnet: 172.26.43.0/24
+      gateway: 172.26.43.1
+
```
anadahz: I guess subnet and gateway should be also to the var/main.yml file

ansible/roles/airflow/tasks/main.yml
```diff
@@ -0,0 +1,155 @@
...
+- name: shared airflow config templates
+  file: dest=/srv/etc/af-share-{{ airflow_tag }} state=directory owner=root group=root mode=0750
+- stat: path=/srv/etc/af-share-{{ airflow_tag }}/passwd
+  register: afpasswd
+- stat: path=/srv/etc/af-share-{{ airflow_tag }}/group
+  register: afgroup
+- stat: path=/srv/etc/af-share-{{ airflow_tag }}/airflow.cfg
+  register: afcfg
+
+- name: create temporary openobservatory/airflow container to extract file
```
anadahz: Is this really necessary or perhaps I don't understand why this should in this way.

ansible/roles/airflow/tasks/main.yml
```diff
@@ -0,0 +1,155 @@
...
+- name: mkdir airflow-dags
+  file: dest=/srv/airflow-dags state=directory owner=root group=root mode=0755
+- name: copy airflow-dags # depends on `rsync`
+  tags: airflow-dags
+  synchronize: src=airflow-dags/ dest=/srv/airflow-dags/ archive=no recursive=yes checksum=yes delete=yes
+  notify: restart airflow
+
+- include: af-container.yml dockname=af-sched  login=afsched  command=scheduler
+- include: af-container.yml dockname=af-websrv login=afwebsrv command=webserver ipv4_address=172.26.43.254
+- include: af-container.yml dockname=af-worker login=afworker command=worker image_ext=-worker
```
anadahz: Perhaps it may be easier to follow if the includes where added as separate tasks to the ansible main tasks file.

anadahz:
Does the current playbook support setting up the docker containers in separate hosts as it may be useful in the long-run as containers require more resources eg: a postgresql container growing up.

anadahz:
There a number of ansible tasks that have no name, it will be preferable all ansible tasks so that the ansible control host running the playbook will be able to determine easier if something went wrong in which task it occured.

anadahz:
Perhaps it makes sense to track the upstream changes of airflow specific files that are being checkout to this playbook, eg: https://github.com/puckel/docker-airflow/blob/master/script/entrypoint.sh

ansible/dom0-bootstrap.yml
darkk: Hmmm... It was required for minimal image some time ago. I'll double-check.

ansible/ansible-version.yml
darkk: `all` is used to match any `--limit` filter, `run_once: true` is used to use any single host from the set and `connection: local` to avoid actual ssh roundtrip.

If `hosts: localhost` is used instead then the version is not actually enforced if `localhost` is not part of play.

ansible/roles/docker_py/tasks/main.yml
darkk: **Seems**, it's not true anymore. I'll double-check though.

ext/hosts
darkk: That's a script to be run at controller host that is your/mine PC (as we agreed to avoid separate controller jumphost with all private keys) that likely has nmap installed :)

ansible/ext-inventory.yml
darkk: `sleep 60` is used for server-side to die cleanly. Maybe `async: 60` with heartbeat options is enough for that.

ansible/ext-inventory.yml
darkk: I think, there will be two hosts for that in different DCs. I consider using infra hosts for that, the host is my own serv for a while, that'll be fixed "a bit later".

ansible/play
darkk: I think that having `ooni-sysadmin.git/ansible/` being explicit PWD for execution is OK.

ansible/roles/airflow/tasks/main.yml
darkk: Why? It's filled in passwd file anyway and existence of `/nonexistent` directory clearly shows that something is going wrong :)

ansible/roles/airflow/files/airflow-dags/canning.py
darkk: It's the last processed date in the schedule. I've not enabled recurring scheduling of the tasks as I was going offline for a while. It'll be commented when pipeline starts ticking daily.

ansible/roles/airflow/tasks/af-container.yml
darkk: `[copy] module copies a file on the local box to remote locations`. This task extracts shipped airflow.cfg from docker image and that happens on server side.

ansible/roles/airflow/tasks/main.yml
darkk: The only easy way to get a file from docker image is to create temporary stopped container.

ansible/roles/docker_py/tasks/main.yml
hellais: I can confirm that they changed their base image in a way that this is no longer a requirement.

ansible/dom0-bootstrap.yml
darkk: That's interesting, GH Debian 8.7 and Ubuntu 16.04 images have `python` and ansible works OK, although I was pretty sure that this `pre_task` is there for a reason.
Seems, my memory is betraying me :-)

ansible/roles/adm/tasks/main.yml
darkk: > sudo group

I done it that way as I prefer uid/gid/groups to be global (makes life easier for programs that use numeric values of IDs, like some versions of rsync and tar), that's why it's per-host _ACL_.

> lineinfile

I prefer replacing whole file instead of `linefile` when it's not too complex as it's makes history of edits more transparent. What's wrong with template files?

ansible/roles/airflow/tasks/main.yml
darkk: I use `include` as a function call to reduce amount of copy-paste. Do you think that dict merging in `af-container.yml` is too ugly?

darkk:
> How one should be running this playbook to test it?

Well, technically it's possible to make it more testable, but I have written it as "executable documentation", so things may be recovered from crash more easily.
I was satisfied with running it with `--step`, verifying that it reports no `changes` tasks on consequent runs and I consider hardcore testing of the playbooks an overkill in this case as I consider communication (across people and across time) to be the main value of these playbooks :)

> Perhaps it makes more sense to add all the POSTGRES_* variables of the airflow role to var/main.yml

I have opposite opinion, IMHO, making meta-step like `POSTGRES_FOO: "{{postgres_foo}}"` for every variable that is not going to be changed in foreseeable future is premature metamisation :)

> docker containers in separate hosts

No. It also does not support swarm, k8s and peta-scale operations :)

ansible/roles/airflow/tasks/af-container.yml
darkk: That's true, but I think that `unless-stopped` is even more convenient.

ansible/roles/airflow/tasks/af-container.yml
anadahz:
```
[copy]
[...]
- remote_src
        If False, it will search for src at originating/master
        machine, if True it will go to the remote/target machine for
        the src. Default is False.
        Currently remote_src does not support recursive copying.
        (Choices: True, False)[Default: no]
[...]
```

ansible/roles/airflow/tasks/main.yml
anadahz: Sure why not! :)

ansible/roles/airflow/tasks/main.yml
anadahz: Well according to Ansible's BCP and role organization and sanity of the people that would like to (re)build a similar setup it will be "nicer" IMHO to have the include in the main as separate tasks.

ansible/roles/airflow/tasks/af-container.yml
darkk: That's true, I was lazy debater :)

Maybe `-a` was the reason. I'll think a bit about it and, if `-a` is not the reason I'll replace it with `copy`.

ansible/ext-inventory.yml
anadahz: Also we should ensure that the user can access the ssh server, since the ssh client configuration file may not include this host or use a different HostName.
Ideally we could use the ansible's  [proxy_command](http://docs.ansible.com/ansible/intro_configuration.html#proxy-command) and [ssh_args](http://docs.ansible.com/ansible/intro_configuration.html#ssh-args) for any specific ssh configuration. This way we will avoid the "hackish" way of [connecting to the jumphost](https://github.com/TheTorProject/ooni-sysadmin/blob/c72f7725b3d65d85c8a094b02d80b7d2f4638054/ansible/ext-inventory.yml#L12-L20] and using a command with [curl instead of the uri module](https://github.com/TheTorProject/ooni-sysadmin/blob/c72f7725b3d65d85c8a094b02d80b7d2f4638054/ansible/ext-inventory.yml#L25)

ansible/dom0-bootstrap.yml
anadahz: :) BTW, this is also useful for people that use a separate virtual environment (ansible_python_interpreter).

ansible/roles/adm/tasks/main.yml
anadahz: Agreed! There is nothing wrong with having multiple template files, I mentioned it in case that we were just using one group for all users that should have sudo access.

ansible/roles/airflow/tasks/main.yml
darkk: Seems, I'm misunderstanding you. Do you suggest expanding these includes into list of tasks, moving beefy part of `main.yml` to separate include or even something else?

ansible/ext-inventory.yml
darkk: I avoid `uri` module and use curl with a proxy for a reason -- I don't want a secret token that may be used to update DNS records to be kept on the disk of jumphost as a temporary file.
I'll add a comment about that.

ansible/roles/airflow/tasks/af-container.yml
anadahz: Note that if we use the `unless-stopped` restart_policy and for some reason we manually stop a container this container will not be brought up upon reboot (since it was in a stopped condition).
The cases that we will need to stop a production container will be very few and for sure we can use other ways (if it's needed) to stop containers and not let allow to auto starting them again.

ext/hosts
anadahz: I would prefer if this is not really running on the control host directly, `sudo nmap` will break many things for people that are in restricted and firewall networks.
Does this script requires access to any/all private keys?

ansible/roles/airflow/tasks/af-container.yml
darkk: Yep. That's exactly I consider desired behavior. `docker stop` is rare and is indication of some disaster when manual intervention happens and I think it's good that explicitly stopped container is not auto-restarted on reboot.

ansible/roles/airflow/tasks/main.yml
anadahz: It seems to a bit of an overkill to create a container "just" to extract a file but sure why not.

ext/hosts
darkk: No. Also, this script does not produce anything but a tab-separated table that can be consumed by human :) Also, `nmap` sometimes does not resolve all PTRs, that's why `gethostbyaddr` is also used.

ansible/roles/airflow/tasks/main.yml
darkk: I agree that it looks like overkill, but I have not found easier solution. Anyway, this image will be used as a layer for other containers, so it's not that bad :)

ansible/roles/airflow/tasks/main.yml
anadahz: I was suggesting a structure similar to: [ansible/roles/ooni-backend/tasks/main.yml](https://github.com/TheTorProject/ooni-sysadmin/blob/22444ffcd9df745a802874970609021dbfa6392c/ansible/roles/ooni-backend/tasks/main.yml)

anadahz:
> I have opposite opinion, IMHO, making meta-step like POSTGRES_FOO: "{{postgres_foo}}" for every variable that is not going to be changed in foreseeable future is premature metamisation :)

I see only 6 variables that need to be added in var/main.yml.

If for any reason we will (or anyone that uses this role) need to change them 3 files and 24 entries in these files will need to be edited and hope that there were no typos made :)

```
# grep --exclude-dir=.git -rc 'POSTGRES_*\|AIRFLOW_*' . | grep -v ':0'
./ansible/roles/airflow/tasks/af-container.yml:3
./ansible/roles/airflow/tasks/main.yml:2
./docker/airflow/entrypoint.sh:19
```

ansible/dom0-bootstrap.yml
darkk: I have successfully beaten the confession out of the memory trying to betray me!
It was `ubuntu:16.04` and `debian:jessie` docker images that I used to check syntax of the playbook with `connection: docker`. :)

ansible/ansible-version.yml
anadahz: > If hosts: localhost is used instead then the version is not actually enforced if localhost is not part of play.

How do you call this? Using `hosts: localhost` is how it worked for me while I was testing the assert task.

darkk: Try `./play --check --limit datacollector.infra.ooni.io deploy-pipeline.yml` -- assertion using `hosts: localhost` will be skipped in that case.

PR #106, opened 2017-04-18
[WIP] Add Prometheus monitoring server

SuperQ:
Add a role for installing a Prometheus monitoring server.

### DIFF
--- a/ansible/roles/blackbox_exporter/defaults/main.yml
+++ b/ansible/roles/blackbox_exporter/defaults/main.yml
@@ -1,0 +1,8 @@
+go_arch: amd64
+
+blackbox_exporter_install_path: /srv/blackbox_exporter
+blackbox_exporter_version: 0.4.0
+blackbox_exporter_download_url: https://github.com/prometheus/blackbox_exporter/releases/download
+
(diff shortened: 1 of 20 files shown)


hellais:
This looks great, thanks for putting it together!

Would it be possible to also add some examples on how we would go abouts adding exporters to one of our components?

I think for the moment the most high priority things to monitor is probably [ooni-backend](https://github.com/TheTorProject/ooni-sysadmin/tree/master/ansible/roles/ooni-backend). What we would like to be able to monitor there is:
* Basic machine metrics (with node_exporter)
* Something to monitor if the actual ooni-backend collector (a HTTP(s) service) is running properly (it runs inside of docker) either via a process monitor or by reaching the HTTP(s) endpoints

SuperQ:
Sure, exporter/metrics examples are pretty easy.  I wanted to start with the server side.

For monitoring nodes, we will need to enumerate all of the nodes, this is easy to do with ansible's host inventory.

For monitoring jobs, we will need to enumerate all of the instances of the app running in docker.  If this is also in some kind of ansible inventory, it should be easy to do.

As for how to get the data out of the ooni-backend, it will be best to integrate one of the Prometheus application libraries[0].

[0]: https://prometheus.io/docs/instrumenting/clientlibs/

hellais:
Note: this is superseeded by #108

darkk:
@SuperQ thanks a lot for helping to bootstrap that

Issue #93, opened 2017-03-15
Monitoring of infrastructure with alerts


hellais:
We currently have some basic monitoring setup here: https://munin.ooni.io/.

However it's currently not sending us alerts when something bad happens, which I guess is an improvement, but still not perfect.

What is needed in order to have us receive alerts when something unexpected happens on the infrastructure? I believe the issue lies in the fact that our `@oo` accounts don't receive the emails unless they are sent from a legit SMTP server.

What about setting up a gmail account for notifications and use that?

Alternatively we could go for https://github.com/TheTorProject/ooni-sysadmin/issues/53, but I suspect that is going to be more laborious and also what happens if email server goes down, who is going to alert us of that?

SuperQ:
I could help with setting up Prometheus alerting.

hellais:
> I could help with setting up Prometheus alerting.

We currently have a munin based monitoring system. How much more complex is it to use something like prometheus?

SuperQ:
Prometheus is a little bit of up-front work, but we have Ansible scripts to deal with most of it.

But good alerting is the primary goal of Prometheus.  So the long-term gains are worth it.

darkk:
@SuperQ have you used blackbox_exporter to do active probing as well? Can you mention any pitfalls with prometheus?

SuperQ:
I can talk about Prometheus at length.

Yes, I have done a number of active blackbox probe setups.

I think the biggest pitfall of Prometheus, is the fact that it is a general-use monitoring system, and has a high degree of flexibility.  This flexibility requires a bit of a learning curve to get used to.

The only other thing that requires a bit of planning is that Prometheus collects data over a number of different HTTP ports.  It requires a bit of planning depending on the network design. Prometheus targets are simple HTTP get endpoints, with no writeable API.  Because of this simple design, security was left "up to the user". The target design for use on private networks, not on public IP hosted servers.

hellais:
Some notes related to issues with sending email notifications:

* Having a relay for `@infra.ooni.io` is okayish, having `@openobservatory.org` is probably not a good idea right now

* Connection to smtp server for sending emails can be unreliable (if smtp server is in AMS and airflow is sending emails from HK) so it's maybe better to have a local email (or inside of the same location) queue and that will handle delivering email.

hellais:
[...]

PR #108, opened 2017-05-05
Feature/monitoring

hellais:
This is based on top of: https://github.com/TheTorProject/ooni-sysadmin/pull/106 and https://github.com/TheTorProject/ooni-sysadmin/pull/105

TODO:

* [x] Slack notifications are not formatted nicely
* [x] A part from the color of the notifcation it's unclear if the event is about a machine being down vs it having come back up (may be more of a problem when we bridge it with IRC)
* [x] ~The check for the bouncer being up I ended up reverting it to a much simpler one, because I couldn't figure out how to get the regexps to work~
* [x] Sign CSR locally (on the controller) instead of on the target host
* [x] Set binary permissions to root
* [x] Use separate user for alertmanager
* [x] Add comments to autogenerated files
* [x] Replace `vault_` prefixed variables
* [x] Verify we can use `root` instead `www-data` as users for certificates
* [x] Drop node_exporter options we don't need (such as SMART monitoring)
* [x] Pin the grafana docker image version to something reasonable

This implements: https://github.com/TheTorProject/ooni-sysadmin/issues/93


ansible/group_vars/all/vars.yml
```diff
@@ -0,0 +1,55 @@
...
+  darkk:
+    login: darkk
+    id: 10003
+    comment: Leonid Evdokimov
+    keys: ["ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDWe9jGACZhSKhCq/jGOe8xKpY8bPSC8MagEiWMCzcHA5520Js4EGzwuYx5h/7sHK4GKWekGzO57xOsgeKACDKo8JppIPt+uJtnwJpVr7EHdnX1hhBMfo0NfV42R+//fZIHMNtGIW31iPg3cQVp9c+GQPsdmuDPbKGLoQvS9dGnF+N6lM+dicEmeBrqbA/t5V1j/rrfpmpTNOWef2WsREFdZBQbgD3rZU1lG0CnxNw6ATnFQgqcC1QsyfWD8e1pmDLByoVuYJ6WXBLKgI86hrMNidxGjzmpAn2q+5cbu5HZXOkmny3TBiP0VxPJJ43PuLIFkdCNh50KRPtTug3o38Oj leon+2016@darkk.net.ru"]
+  nuke:
+    login: nuke
+    id: 10004
+    comment: Lorenzo Primiterra
+    keys: ["ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQC+QtDwA/s8vw7aETWGjItUBmgCXgm9jTEjCh/5OofGJ/Bk4kjZ+a6NisUNJRvrySfr89oeJz7VOuHgHTCM1g2CKANAiidcBkepHjdBlHDElV1WBlUxdgl0RsbM4ue8zUxtrmzEQaxLxL769ZHR+gI8PoZau1cSyKG5jVge91Bt1+wvV3dMGvxibM144+uHCrxHpBsbRd9irIgk+vNJmecKn3bSh3aKo9txBjIqBprBrZ1Dqj0x9IKVhD59URcO1Vrr3HXPghZnsWFjOSSQ0DeVpkE3cEz78CpdvqL0drv4Mo1ZgE33YLwWjJSceDGIrtnsPZhs5YzOI4iStRIdPD4X nuke"]
```
darkk: /etc/passwd at prometheus.infra.ooni.io contains
`prometheus:x:10004:10004:Prometheus User:/home/prometheus:/sbin/nologin`

hellais: The issue you see is that the uid is conflicting?

Do you see a way of avoiding possible conflicts with existing UID?

Maybe we want to put the UIDs for our usernames in a UID space that is not likely to conflict? Maybe offset all our UIDs by 100 or 200.

SuperQ: I'm not sure how this variable gets used, but the [user](http://docs.ansible.com/ansible/user_module.html) module lets you set `system: yes` on accounts.  I would suggest using this, and letting the system automatically pick random UID's in the system range for the `prometheus` user.

darkk: > random UID

They are not random but sequential. That's why I'm adept of static UIDs church :) It becomes useful when some tool does not resolve user names, but uses raw numeric uids (e.g. NFS, old versions of tar & rsync) and files are transferred across machines.

SuperQ:
This PR is getting pretty big!

hellais:
I rebased this on top of master to hopefully improve readability.

ansible/roles/blackbox_exporter/templates/blackbox_exporter.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=Prometheus Blackbox Exporter
+
+[Service]
+User=nobody
+Group=nogroup
+EnvironmentFile=-/etc/default/blackbox_exporter
+PermissionsStartOnly=true
+ExecStartPre=/sbin/setcap 'cap_net_raw+ep' {{ blackbox_exporter_path }}/blackbox_exporter
```
darkk: Why does it need `cap_net_raw`? I **assume**, it's for ICMP PING, but it can be done with SOCK_DGRAM, IPPROTO_ICMP as well.

hellais: Yeah it's needed for ICMP ping as well as for the DNS based probes. The docs of blackbox_exporter mention these as the [recommended settings](https://github.com/prometheus/blackbox_exporter#permissions).
Maybe @SuperQ knows more.

SuperQ: It's only needed for ICMP.  DNS is done through normal resolver methods.

The actual implementation is based on the upstream Go code: https://godoc.org/golang.org/x/net/icmp

It looks like this used to not work, but should work now: https://github.com/golang/go/issues/9166

We should test this with minimal privs. If it works, I can update the upstream docs.

SuperQ: According to http://man7.org/linux/man-pages/man7/capabilities.7.html, there seems to be no other way to grant `SOCK_DGRAM`, `IPPROTO_ICMP`.

darkk: AFAIK, IPPROTO_ICMP is granted using EGID via `net.ipv4.ping_group_range` sysctl on Linux (group vector is not enough), I know nothing about MacOSX.
I suggest to continue the discussion at prometheus/blackbox_exporter#147

ansible/roles/adm/tasks/main.yml
```diff
@@ -39,3 +39,10 @@
 
 - name: sudoers.d/adm
   template: src=sudoers dest=/etc/sudoers.d/adm owner=root group=root mode=0440 validate='visudo -cf %s'
+
+- name: install sudo
+  apt:
+    name: "{{ item }}"
+    state: latest
+  with_items:
+    - "vim"
```
darkk: I'd suggest to merge it with `install sudo` in the beginning of the file as `apt` module has special fast-path to install all the packages at once.

ansible/roles/alertmanager/defaults/main.yml
```diff
@@ -0,0 +1,24 @@
+go_arch: amd64
+
+alertmanager_user:   prometheus
+alertmanager_group:  prometheus
```
darkk: There is no need to share UID between alertmanager and prometheus daemons.

ansible/roles/alertmanager/tasks/main.yml
```diff
@@ -0,0 +1,75 @@
...
+    - "{{ alertmanager_templates_path }}"
+
+- name: untar alertmanager tarball
+  unarchive:
+    src: "{{ alertmanager_tarball_url }}"
+    dest: "{{ alertmanager_install_path }}"
+    mode: 0755
+    owner: "{{ alertmanager_user }}"
+    group: "{{ alertmanager_group }}"
+    copy: no
```
darkk: It is considered questionable practice when the daemon owns its own binary, binaries are usually owned by root if UID separation is used.

ansible/roles/grafana/tasks/main.yml
```diff
@@ -0,0 +1,29 @@
...
+  docker_container:
+    image: grafana/grafana
+    name: grafana
+    ports:
+      - "{{ grafana_listen_address }}:3000"
+    volumes:
+      - "{{ grafana_config_path }}:/etc/grafana"
+      - "{{ grafana_lib_path }}:/var/lib/grafana"
+    env:
+      GF_SECURITY_ADMIN_PASSWORD: "{{ grafana_admin_password }}"
```
darkk: I'd prefer version being explicitly stated.

ansible/roles/grafana/tasks/main.yml
```diff
@@ -0,0 +1,29 @@
+---
+- name: mkdir for config and data
+  file:
+    path: "{{ item }}"
+    state: directory
+    mode: "u=rwx,g=rx,o="
+  with_items:
+    - "{{ grafana_config_path }}"
+    - "{{ grafana_lib_path }}"
```
darkk: Files in /srv/grafana/ are owned by `sshd:crontab` at the moment. Is it possible to allocated separate UID for the service or does `grafana/grafana` need to be started as root?

ansible/roles/grafana/tasks/setup-nginx.yml
```diff
@@ -0,0 +1,31 @@
+---
+- name: Install nginx
+  environment:
+    RUNLEVEL: 1
```
darkk: IMHO, it deserves a comment like `# to avoid nginx being launched right after installation`

ansible/roles/grafana/templates/grafana_nginx.conf.j2
```diff
@@ -0,0 +1,20 @@
+server {
```
darkk: I usually write a comment like `# ansible-managed in ooni-sysadmin.git` in every managed file so it's obvious that it should not be edited manually.
I was inspired by lots of file generators adding `DO NOT EDIT! File autogenerated by blah-blah.` comment :)

ansible/roles/grafana/templates/grafana_nginx.conf.j2
```diff
@@ -0,0 +1,20 @@
...
+    server_name _;
+
+    listen 3000 ssl default_server;
+    listen [::]:3000 ssl default_server;
```
darkk: Is there a reason for it to be `:3000` and not `:443` ?

ansible/roles/node_exporter/defaults/main.yml
```diff
@@ -0,0 +1,11 @@
+go_arch: amd64
+
+node_exporter_install_path: /srv/node_exporter
+node_exporter_textfiles_path: /srv/node_exporter/textfiles
+node_exporter_collectors: diskstats,filefd,filesystem,loadavg,mdadm,meminfo,netdev,netstat,sockstat,stat,textfile,time,uname
```
darkk: Do we have anything managed by `mdadm` ?

ansible/roles/node_exporter/tasks/main.yml
```diff
@@ -0,0 +1,41 @@
...
+    daemon_reload: yes
+
+- name: Install node_exporter defaults file
+  template:
+    src: defaults
+    dest: /etc/default/node_exporter
+  notify:
+  - restart node_exporter
+
+- include: smartmon_metrics.yml
```
darkk: Do we have any bare-metal servers we're not migrating away from? As far as I understand, S.M.A.R.T. is useless for VMs.

ansible/roles/node_exporter/templates/node_exporter.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Prometheus Node Exporter
+
+[Service]
+User=nobody
+Group=nogroup
```
darkk: LSB suggests using distinct UIDs instead of shared `nobody`.

ansible/roles/node_exporter_nginx/defaults/main.yml
```diff
@@ -0,0 +1,9 @@
+prometheus_basic_auth_user: "prometheus"
+prometheus_basic_auth_password: "{{ vault_prometheus_basic_auth_password }}"
+node_exporter_user: "www-data"
+node_exporter_group: "www-data"
+node_exporter_private_path: "/srv/node_exporter/private"
+node_exporter_ssl_key_path: "{{ node_exporter_private_path }}/{{ inventory_hostname }}.pem"
+node_exporter_ssl_csr_path: "{{ node_exporter_private_path }}/{{ inventory_hostname }}.csr"
+node_exporter_ssl_cert_path: "{{ node_exporter_private_path }}/{{ inventory_hostname }}.cert"
```
darkk: IIRC, `nginx` (re)loads all ssl secrets before dropping `root` privileges. Is it the only reason to use `www-data` for certs?

ansible/roles/node_exporter_nginx/tasks/create-certs.yml
```diff
@@ -0,0 +1,72 @@
...
+- name: Set CSR permissions
+  file:
+    path: "{{ node_exporter_ssl_csr_path }}"
+    owner: "{{ node_exporter_user }}"
+    group: "{{ node_exporter_group }}"
+    mode: "0700"
+
+- name: write the root ca private key to temporary directory
+  copy:
+    content: "{{ vault_prometheus_root_ca_key }}"
```
darkk: The idea behind `vault_` prefix and defining variables like `foo: "{{ vault_foo }}"` is to show exact location of vaulted variable as vaults are not grepable. So `vault_*` variables should not be used anywhere besides `vars.yml`.

ansible/roles/node_exporter_nginx/tasks/create-certs.yml
```diff
@@ -0,0 +1,72 @@
...
+    dest: "{{ node_exporter_private_path }}/root_ca.key"
```
darkk: FYI: `copy` supports `mode` option as well (maybe I'm wrong and it's broken when `content` is used).

But I still suggest signing CSR locally at controller host.

ansible/roles/node_exporter_nginx/tasks/create-certs.yml
```diff
@@ -0,0 +1,72 @@
...
+  command: openssl x509 -req -sha256 -days 1024 -in {{ node_exporter_ssl_csr_path }} -CA "{{ node_exporter_private_path }}/root_ca.pem" -CAkey "{{ node_exporter_private_path }}/root_ca.key" -CAcreateserial -out {{ node_exporter_ssl_cert_path }} -extensions v3_ca creates={{ node_exporter_ssl_cert_path }}
+
+- name: Set signed cert permissions
+  file:
+    path: "{{ node_exporter_ssl_cert_path }}"
+    owner: "{{ node_exporter_user }}"
+    group: "{{ node_exporter_group }}"
+    mode: "0700"
+
+# XXX there should be a way to delete this even if the previous steps fail
```
darkk: `notify` is probably yet another imperfect way to do it :)

ansible/roles/node_exporter_nginx/templates/node_exporter_nginx.conf.j2
```diff
@@ -0,0 +1,28 @@
...
+  location / {
+    auth_basic "Monitoring realm";
+    auth_basic_user_file /etc/nginx/prometheus.htpasswd;
+
+    proxy_pass http://{{ node_exporter_listen_address }};
+    proxy_http_version 1.1;
+    proxy_set_header Host $http_host;
+    proxy_set_header X-Real-IP $remote_addr;
+    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+    proxy_set_header X-Forwarded-Proto $scheme;
```
darkk: Does node_exporter actually uses any of these headers?...

ansible/roles/prometheus/tasks/main.yml
```diff
@@ -0,0 +1,77 @@
...
+    - "{{ prometheus_db_path }}"
+
+- name: untar prometheus tarball
+  unarchive:
+    src: "{{ prometheus_tarball_url }}"
+    dest: "{{ prometheus_install_path }}"
+    mode: 0755
+    owner: "{{ prometheus_user }}"
+    group: "{{ prometheus_group }}"
+    copy: no
```
darkk: Same reasoning about daemon owning the binary and being able to write to it.

.travis.yml
```diff
@@ -0,0 +1,7 @@
+language: python
+python:
+  - "2.7"
+install:
+  - pip install ansible==2.2.2
+script:
+  - cd ansible && ansible-playbook --version
```
hellais: Yes that is true, maybe we can, however, have a special vault with dummy values that is only used in CI and at least runs the syntax check.

ansible/deploy-node-exporter.yml
```diff
@@ -0,0 +1,20 @@
...
+#        - "{{ passwd.art }}"
+#        - "{{ passwd.darkk }}"
+
+- hosts: monitored
+  gather_facts: true
+  vars:
+    node_exporter_listen_address: "127.0.0.1:9001"
+  roles:
+    - node_exporter
+    - node_exporter_nginx
```
hellais: I agree, however I would suggest doing that only after we have tested that everything works as expected on a handful of nodes. At that point we should re-run dom0 with `targets: all`.

ansible/host_vars/prometheus.infra.ooni.io/vars.yml
```diff
@@ -0,0 +1,40 @@
...
+  - name: "ooni bouncer"
+    module: "ooni_bouncer_ok"
+    targets:
+      - "https://bouncer.ooni.io/bouncer"
+
+  - name: "ooni tcp echo"
+    module: "ooni_tcp_echo_ok"
+    targets:
+      - "37.218.247.110:80"
+      - "213.138.109.232:80"
```
hellais: Because this is the value that is returned by the bouncer. I would rather we have checks that verify things as if it were a real ooniprobe to avoid possible issues.

.travis.yml
darkk: Yep, that's possible, but it needs some vault variables listing method. Accurate `vault_` prefix manipulation and stripping the prefix in corresponding `vars.yml` should work :)

ansible/host_vars/prometheus.infra.ooni.io/vars.yml
```diff
@@ -0,0 +1,40 @@
...
+  - name: "ooni http return json headers"
+    module: "ooni_http_return_json_headers_ok"
+    targets:
+      - "http://38.107.216.10:80"
```
hellais: This is the default http_return_json_headers test-helper set in the bouncer: `http-return-json-headers: 'http://38.107.216.10:80'`.
From the IP it looks like it's running in the WDC datacenter, but I also cannot find it in nor DNS records, nor in the eclips.is frontend.

@anadahz do you know where this is running from?

ansible/roles/adm/tasks/main.yml
hellais: Ok makes sense.

ansible/roles/alertmanager/defaults/main.yml
hellais: Fair enough, I guess we can create a special user for alertmanager in that case.

ansible/roles/grafana/tasks/main.yml
hellais: Grafana currently runs inside of a docker container. I am not sure what is the best way to setup permissions properly for this. In theory since we map volumes the grafana process should actually not have access to anything else.

ansible/roles/grafana/tasks/main.yml
hellais: Version of what?

darkk: I mean `grafana/grafana` image.

ansible/roles/grafana/templates/grafana_nginx.conf.j2
hellais: Not really. The only issue I can see with putting it on 443 is that we need to do some more magic to get letsencrypt to work properly (take down grafana docker, start it back up again upon renewal).

ansible/roles/node_exporter/defaults/main.yml
hellais: I don't think so. This was part of the playbook that @SuperQ provided for node_exporter.

ansible/roles/node_exporter/tasks/main.yml
hellais: Again this is also something that was part of the roles @SuperQ provided, I didn't change it that much, but yeah I guess you are right, we can probably drop this.

ansible/roles/node_exporter_nginx/defaults/main.yml
hellais: Ha, ok. I wasn't aware of that actually. That is in fact the only reason to set those as www-data. If we confirm that is the case we can do that.

ansible/roles/node_exporter_nginx/tasks/create-certs.yml
hellais: Yeah, this is something I think I wrote before I figured out that was the pattern you were using.

ansible/host_vars/prometheus.infra.ooni.io/vars.yml
darkk: It looks like some box in cogento networks, but not GH WDC. RTT also suggests quite different locations.

ansible/roles/node_exporter_nginx/templates/node_exporter_nginx.conf.j2
hellais: I seem to recall reading somewhere that was good practice of headers to include (can't find link now though). In any case I think that it's good to forward these headers as in theory we should be the only ones accessing it and maybe there is some way to set it up to not respond to probes from unknown IPs.

SuperQ: `node_exporter` doesn't care about those.  The only thing it really wants to know is the encoding headers so it can correctly respond with text or protobuf.  Also compression encoding.

ansible/roles/grafana/tasks/main.yml
darkk: docker processes share UID namespace with "dom0", that's why I have separate UIDs in dom0 for every airflow service. It also requires some `passwd` file patching in the image, but I consider it useful for two reasons:
- `ps`, `top` and alike utils have ability to filter out processes for specific UID. Having all dockerized processes running under uid#1000 (named differently in every container) breaks this pattern.
- `ls -l` does not show anything creepy in dom0

ansible/host_vars/prometheus.infra.ooni.io/vars.yml
hellais: Mumble. In that case I wonder if maybe it's an mlab node. If that is the case we should probably switch over to using one of the ones that live on greenhost and is deployed by us.

ansible/roles/grafana/tasks/main.yml
hellais: Ah right. Yes I agree we should probably pin it to something reasonable.

ansible/roles/grafana/tasks/main.yml
hellais: I am going for the moment defer this to future work and not consider it part of MVP.

ansible/roles/grafana/tasks/main.yml
hellais: Done.

ansible/roles/grafana/tasks/setup-nginx.yml
hellais: Done/

ansible/roles/grafana/templates/grafana_nginx.conf.j2
hellais: Done.

ansible/roles/grafana/templates/grafana_nginx.conf.j2
hellais: I will defer this to after MVP.

ansible/roles/node_exporter/defaults/main.yml
hellais: Done.

ansible/roles/node_exporter/templates/node_exporter.service
hellais: Done.

ansible/roles/prometheus/tasks/main.yml
hellais: Done.

hellais:
I moved the bouncer health check improvement to another ticket: https://github.com/TheTorProject/ooni-sysadmin/issues/113 and shall defer it to future work.

For MVP purposes I think this is ready to be merged and deployed.

hellais:
Can somebody take a look at this PR? I think it's ready to be merged.

darkk:
Let me fix UID collision before merging.

hellais:
> Let me fix UID collision before merging

The UID collision is now fixed. In fixing it I noticed a bunch of other places where the UID collision is still happening (for example roles for ooni-explorer). I would fix that however in another pass (i.e. pull request) as this one is already getting a bit large.

hellais:
Think this is ready to land in master and be deployed?

.travis.yml
darkk: [It](https://github.com/TheTorProject/ooni-sysadmin/commit/9b48318dd8029f39b5fb834ce7b8147ec9b8172c) works! https://travis-ci.org/TheTorProject/ooni-sysadmin/builds/257356900

darkk:
Yay! https://travis-ci.org/TheTorProject/ooni-sysadmin/builds/260663666

I'm merging this to prevent further divergence between branches and further cleanup and exporter deployments will (eventually) happen at #132

Issue #101, opened 2017-03-29
common ansible template


darkk:
Writing it down to share (and be stopped if anyone has objections):

- [ ] NTP
- [ ] `root` mail going somewhere instead of being dropped on the floor
- [ ] shell and `sudo` for @anadahz @hellais @darkk #92 
- [ ] **nearby** debian repo, to avoid going from HKG to `ftp.nl.debian.org`
- [ ] static network configuration in `interfaces` (with firewall, if necessary), `resolv.conf` with default GH setting (195.190.28.{82,85} in AMS, OpenDNS in HKG), `hosts` with local FQDN
- [ ] `sshd_config` with password-based access disabled
- [ ] latest `tzdata`
- [ ] deb: rsync, wget, tmux, lsof, strace, tcpdump, mtr, iptables, dnsutils (dig), bind9-host (host), atop
- [ ] alerts for: ntp, disk space usage (df), memory usage #93

SuperQ:
I can recommend chrony[0] as a good NTP client.  It has a bunch of [advantages](https://chrony.tuxfamily.org/comparison.html).

[0]: https://packages.debian.org/search?keywords=chrony

darkk:
@SuperQ thanks, I also lean towards simplistic openntpd or chrony.

PR #112, opened 2017-06-12
Feature/orchestration

hellais:
This is based on top of the monitoring pull request and includes all the roles for deploying probe orchestration backends.


bassosimone:
Oh, my, how many conflicts!

ansible/deploy-proteus-frontend.yml
```diff
@@ -0,0 +1,26 @@
+---
+- hosts: proteus.test.ooni.io
+  vars:
+    proteus_frontend_https_port: 443
+    proteus_events_url: "https://events.proteus.test.ooni.io"
+    proteus_registry_url: "https://registry.proteus.test.ooni.io"
+    proteus_version: "v0.1.0-beta.4"
```
bassosimone: @hellais for uniformity, here you may want to _bump_ to the latest version (AFAICT, nothing changed).

From my notes, I also noticed two bugs in the frontend that I was not able to fix:

- the view of message tasks to be sent to users triggered a JavaScript exception when no tasks were registered

- the list of registered probes and message tasks does not seem to be sorted according to time.

I guess these comments can be translated into proper tickets for proteus.

ansible/deploy-proteus-frontend.yml
```diff
@@ -0,0 +1,26 @@
...
+    - letsencrypt
+    - nodejs
+    - proteus-frontend
+
+- hosts: proteus.ooni.io
+  vars:
+    proteus_frontend_https_port: 443
+    proteus_events_url: "https://events.proteus.ooni.io"
+    proteus_registry_url: "https://registry.proteus.ooni.io"
+    proteus_version: "v0.1.0-beta.4"
```
bassosimone: ditto

ansible/deploy-proteus-notify.yml
```diff
@@ -0,0 +1,31 @@
...
+  vars:
+    proteus_notify_https_port: 443
+    letsencrypt_nginx: yes
+    letsencrypt_domains: "{{ inventory_hostname }}"
+    proteus_database_url: "postgres://proteus:{{ vault_proteus_database_password_testing }}@db-1.proteus.test.ooni.io:5432/proteus?sslmode=require"
+    proteus_auth_jwt_secret : "{{ vault_proteus_auth_jwt_secret_testing }}"
+    proteus_fcm_server_key: "{{ vault_fcm_server_key_testing }}"
+    proteus_apn_key_password: "{{ vault_proteus_apn_key_password_testing }}"
+    proteus_apn_key_content: "{{ vault_proteus_apn_key_content_testing }}"
+    proteus_environment: "production"
```
bassosimone: Just checking here: it has always been `production` and everything worked for us, but I've never been sure whether this is the correct setting or not. What is the meaning of this key?

ansible/roles/adm/tasks/main.yml
```diff
@@ -4,6 +4,10 @@
 # point of truth for the keys, unknown keys are dropped, that's why
 # `authorized_key` ansible module is not used.
 
+- name: Run "apt-get update"
+  apt:
+    update_cache: yes
+
```
bassosimone: I guess this is the place for a more general question for @darkk and @hellais: how do we manage updates on the remote machines? My limited understanding is that, as of now, we just `apt-get update` when we deploy new stuff, is that correct? Do we turn on `unattended-upgrades` or something similar for _security_ updates?

ansible/roles/alertmanager/tasks/main.yml
```diff
@@ -0,0 +1,75 @@
+- name: create Alertmanager group
```
bassosimone: My understanding is that these strings are case sensitive and that this may have implications especially when we want to restart things. So, maybe, it would make sense to always use lowercase?

ansible/roles/postgresql/tasks/main.yml
```diff
@@ -3,4 +3,4 @@
 - include: config.yml
   become: yes
   become_user: postgres
-  become_method: su
+  become_method: sudo
```
bassosimone: We do have `sudo` in `dom0`, right?

ansible/roles/nodejs/tasks/install-yarn.yml
```diff
@@ -0,0 +1,12 @@
+---
+- name: Add YarnPkg apt package signing key
+  apt_key:
+    url: "https://dl.yarnpkg.com/debian/pubkey.gpg"
```
darkk: It's "better" to import from keyservers using full fingerprint.

ansible/roles/nodejs/tasks/install-yarn.yml
```diff
@@ -0,0 +1,12 @@
...
+
+- name: Add YarnPkg apt repository
+  apt_repository:
+    repo: 'deb https://dl.yarnpkg.com/debian/ stable main'
```
darkk: Is `stable` for stable yarn release or for debian:stable ?

ansible/roles/postgresql/tasks/main.yml
darkk: Right, Also, that' default become_method.

ansible/roles/proteus-db/tasks/main.yml
```diff
@@ -0,0 +1,13 @@
...
+    src: "{{ item }}"
+    dest: "{{ proteus_db_schema_path }}"
+    mode: 0644
+  with_items: "{{ proteus_db_files }}"
+
+- name: Run database schema commands
+  shell: "psql -U {{ proteus_db_user}} -h 127.0.0.1 -d {{ proteus_db_name }} -f {{ proteus_db_schema_path}}/{{ item }}"
+  environment:
+    PGPASSWORD: "{{ proteus_db_password }}"
+  with_items: "{{ proteus_db_files }}"
```
darkk: What will happen when one needs to ALTER something?

ansible/roles/proteus-events/tasks/main.yml
```diff
@@ -0,0 +1,63 @@
...
+  group: name={{ proteus_group }}  state=present
+
+- name: create Proteus user
+  user:
+    name: "{{ proteus_user }}"
+    group: "{{ proteus_group }}"
+    createhome: no
+    shell: /sbin/nologin
+    comment: "Proteus User"
+    state: present
```
darkk: `/sbin/nologin` is not a valid shell.

UID will clash with superq as soon as dom0 with node_exporter is deployed there:
```
events.proteus.test.ooni.io | SUCCESS | rc=0 | (stdout) uid=10007(proteus) gid=10007(proteus) groups=10007(proteus)
notify.proteus.test.ooni.io | SUCCESS | rc=0 | (stdout) uid=10007(proteus) gid=10007(proteus) groups=10007(proteus)
proteus.test.ooni.io | SUCCESS | rc=0 | (stdout) uid=10007(proteus) gid=10007(proteus) groups=10007(proteus)
registry.proteus.test.ooni.io | SUCCESS | rc=0 | (stdout) uid=10007(proteus) gid=10007(proteus) groups=10007(proteus)
notify.proteus.ooni.io | SUCCESS | rc=0 | (stdout) uid=10006(proteus) gid=10006(proteus) groups=10006(proteus)
registry.proteus.ooni.io | SUCCESS | rc=0 | (stdout) uid=10006(proteus) gid=10006(proteus) groups=10006(proteus)
events.proteus.ooni.io | SUCCESS | rc=0 | (stdout) uid=10006(proteus) gid=10006(proteus) groups=10006(proteus)
proteus.ooni.io | SUCCESS | rc=0 | (stdout) uid=10006(proteus) gid=10006(proteus) groups=10006(proteus)
```

ansible/roles/proteus-events/templates/proteus-events.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Proteus Events service
+
+[Service]
+User={{ proteus_user }}
+Group={{ proteus_group }}
+ExecStart={{ proteus_path }}/proteus-events --config {{ proteus_config_path }}/proteus-events.toml start
```
darkk: Is it the user for `proteus-events` or for `proteus-frontend` ?

ansible/roles/proteus-notify/templates/gorush.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=GoRush Notify service
+After=network.target
+
+[Service]
+User={{ proteus_user }}
+Group={{ proteus_group }}
```
darkk: Is it the user for proteus-frontend or for gorush ?

ansible/deploy-orchestration.yml
```diff
@@ -0,0 +1,33 @@
+---
+- include: ansible-version.yml
+
+- hosts: orchestration
+  gather_facts: true
+  roles:
+    - docker_py
```
darkk: The single use of `orchestration` group I see is to install docker and docker is useless for `db-1` hosts. What's the point of that group?

ansible/deploy-orchestration.yml
```diff
@@ -0,0 +1,33 @@
...
+
+- hosts: db-1.proteus.test.ooni.io
+  gather_facts: false
+  vars:
+    psql_db_name: proteus
+    psql_db_user: proteus
```
darkk: It's good to have different db_name and db_user for grepability

ansible/deploy-proteus-events.yml
```diff
@@ -0,0 +1,27 @@
+- hosts: events.proteus.test.ooni.io
+  vars:
+    proteus_events_https_port: 443
+    letsencrypt_nginx: yes
+    letsencrypt_domains: "{{ inventory_hostname }}"
+    proteus_database_url: "postgres://proteus:{{ proteus_database_password_testing }}@db-1.proteus.test.ooni.io:5432/proteus?sslmode=require"
```
darkk: `sslmode=require` is nice, but it's using `/etc/ssl/certs/ssl-cert-snakeoil.`, so what's the point?

ansible/roles/proteus-events/tasks/main.yml
```diff
@@ -0,0 +1,63 @@
...
+    - "{{ proteus_config_path }}"
+
+- name: untar proteus tarball
+  unarchive:
+    src: "{{ proteus_tarball_url }}"
+    dest: "{{ proteus_install_path }}"
+    mode: 0755
+    owner: "{{ proteus_user }}"
+    group: "{{ proteus_group }}"
+    copy: no
```
darkk: @bassosimone  please, add sha256sums file to releases of proteus binaries and the script that generates releases.

These checksum files may be used to speedup re-run of the playbook as `ger_url` module accepts `checksum` argument [like that](https://github.com/TheTorProject/ooni-sysadmin/blob/ca26b09c46d3c36006f77cbaa866f00654977970/ansible/roles/node_exporter/tasks/main.yml#L19).

ansible/deploy-proteus-notify.yml
```diff
@@ -0,0 +1,31 @@
...
+    proteus_notify_https_port: 443
+    letsencrypt_nginx: yes
+    letsencrypt_domains: "{{ inventory_hostname }}"
+    proteus_database_url: "postgres://proteus:{{ proteus_database_password_testing }}@db-1.proteus.test.ooni.io:5432/proteus?sslmode=require"
+    proteus_auth_jwt_secret : "{{ proteus_auth_jwt_secret_testing }}"
+    proteus_fcm_server_key: "{{ fcm_server_key_testing }}"
+    proteus_apn_key_password: "{{ proteus_apn_key_password_testing }}"
+    proteus_apn_key_content: "{{ proteus_apn_key_content_testing }}"
+    proteus_environment: "production"
+    proteus_notify_basic_auth_password: "{{ proteus_notify_basic_auth_password_testing }}"
```
darkk: @hellais and @bassosimone, please, read paragraph on [variable precedence in ansible](http://docs.ansible.com/ansible/latest/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable), it explains that you _probably_ don't need inderection like `proteus_database_password_testing`, you just need `proteus_database_password` defined in right places

ansible/roles/proteus-events/tasks/main.yml
darkk: UID clash is a blocker to merge it back to master, so I'll have to fix it and redeploy. It'll require some minor proteus downtime.

ansible/roles/nodejs/tasks/install-yarn.yml
hellais: It's for debian:stable

ansible/roles/nodejs/tasks/install-yarn.yml
hellais: Makes sense.

ansible/roles/proteus-db/tasks/main.yml
hellais: Actually all of this stuff can probably be removed as it's now part of proteus itself. We only need to create the proteus_db_user and proteus will take care of populating the database.

ansible/roles/proteus-events/tasks/main.yml
hellais: File a ticket about it: https://github.com/TheTorProject/proteus/issues/22

ansible/roles/proteus-events/templates/proteus-events.service
hellais: They all use the same user (proteus)

ansible/roles/proteus-notify/templates/gorush.service
hellais: In this case it's for gorush.

ansible/roles/proteus-events/templates/proteus-events.service
hellais: In any case this is the user running the proteus-events binary.

ansible/deploy-orchestration.yml
hellais: I think I was previously using this group as a "hack" to run dom0 and gather_facts pre-emptively. I guess it can be removed.

ansible/deploy-proteus-events.yml
hellais: It's better to protect against a purely passive attack and requiring an active attack to leak the password.
Also it's probably not a big deal since we are communicating inside of our network.

ansible/deploy-proteus-notify.yml
hellais: If you can clean it up a bit that would be good. Though while deploying a first version of production I accidentally used the testing password in production and to avoid the risk of confusion I found it more clear to just label them as such.

bassosimone: > Though while deploying a first version of production I accidentally used the testing password in production and to avoid the risk of confusion I found it more clear to just label them as such.

Yeah, I guess this is a good argument in favour of a little more redundancy and robustness.

ansible/roles/proteus-events/templates/proteus-events.service
darkk: I mean, is there any reason for these services to share unix user? As far as I see, there is no -- these services share no objects.

I'm sorry for the form of the question :-)

ansible/roles/proteus-db/tasks/main.yml
darkk: IIRC, @bassosimone was struggling with sql updates embedded into the binary while launching concurrent processes as same DB is used by various services. So there was a suggestion to move SQL files away from the binary and apply them using some other tooling. Is it still the case?

ansible/roles/proteus-events/tasks/main.yml
darkk: Minutes if everything is OK, couple of hours if something goes seriously wrong :)

ansible/roles/proteus-db/tasks/main.yml
hellais: No, I think you were the one suggesting that as a solution, but the issue in the migration scripts was solved, so there is no plan to go that route in the near future.

ansible/roles/proteus-events/templates/proteus-events.service
hellais: No reason.

ansible/deploy-proteus-notify.yml
darkk: > robustness

Yep. I just wanted to say, that reuse of `proteus_user` in four different roles is probably a bad idea :-)

ansible/roles/proteus-notify/defaults/main.yml
```diff
@@ -0,0 +1,37 @@
...
+
+proteus_tarball_url: "{{ proteus_download_url }}/v{{ proteus_version }}/proteus-notify-{{ proteus_version }}.linux-{{ go_arch }}.tar.gz"
+proteus_path: "{{ proteus_install_path }}/proteus-notify-{{ proteus_version }}.linux-{{ go_arch }}"
+
+gorush_version: "1.9.0"
+gorush_download_url: "https://github.com/appleboy/gorush/releases/download"
+gorush_url: "{{ gorush_download_url }}/v{{ gorush_version }}/gorush-v{{ gorush_version }}-linux-{{ go_arch }}"
+gorush_path: "{{ proteus_install_path }}/gorush-v{{ gorush_version }}-linux-{{ go_arch }}"
+
+notification_backend: "gorush"
```
darkk: Should code for `notification_backend: proteus` be removed and corresponding binaries and services cleaned up?

ansible/roles/proteus-registry/handlers/main.yml
```diff
@@ -0,0 +1,4 @@
+- name: restart proteus
+  systemd: name=proteus-registry.service state=restarted
+- name: reload proteus
+  systemd: name=proteus-registry.service state=reloaded
```
darkk: IMHO same handler name for different actions is a way to subtle bugs.

ansible/roles/adm/tasks/main.yml
darkk: > how do we manage updates

We manage it with power of `nop`. I mean, we don't. :-( I add another bullet to #101

ansible/roles/alertmanager/tasks/main.yml
darkk: IMHO, it should be enforced only for _handlers_ and other tasks should have `name` written with preference given to _readability_.

PR #89, opened 2017-02-22
Add slackin ansible role, fix docker pip version bugs

anadahz:

PR #114, opened 2017-06-24
re-integrate role/slackin

darkk:
#89 part 2


hellais:
Ok this is now actually ready for review and being merged.

darkk:
> ready for review

The lightning strikes twice :-)

darkk:
There is also an option in old slack-irc-config.json `"muteBots": ["TOKENISH-STRING"]`, does anyone remember why the options was there? @hellais @anadahz

PR #125, opened 2017-07-17
Deploy measurements-beta.ooni.io

darkk:
Comments are welcome!

I'll be back online on the 24th (hopefully).


ansible/group_vars/all/vars.yml
```diff
@@ -9,6 +9,7 @@ passwd:
   afflower: {login: afflower, id: 2005, comment: Airflow Celery Flower}
   plpsql:   {login: plpsql,   id: 2100, comment: Pipeline Postgresql}
   plshovel: {login: plshovel, id: 2101, comment: Pipeline Shovel}
+  oomsmweb: {login: oomsmweb, id: 2300, comment: OONI Measurements WebAPI}
```
hellais: I snagged this UID in some other deployment I think.

ansible/roles/ooni-measurements/tasks/main.yml
```diff
@@ -1,19 +1,29 @@
...
+- name: ooni-measurements webservice
+  docker_container:
+    image: openobservatory/ooni-measurements:{{ oomsm_tag }}
+    name: oomsm-web
+    hostname: oomsm-web
+    networks: [{name: msm, ipv4_address: '{{ oomsm_backend_ipv4 }}'}]
+    purge_networks: true
+    env:
+      APP_ENV: production
+      PRODUCTION: 'TRUE' # FIXME: is it ever used?
```
hellais: Yes this is used by flask to disable debugger and other development features.

ansible/roles/ooni-measurements/tasks/main.yml
```diff
@@ -1,19 +1,29 @@
...
+      DATABASE_URL: 'postgresql://oomsm-beta:{{ oomsm_beta_postgres_password }}@hkgmetadb.infra.ooni.io/metadb' # FIXME: ?sslmode=require
+      AUTOCLAVED_BASED_URL: 'http://datacollector.infra.ooni.io/ooni-public/autoclaved/' # FIXME: drop this as soon as code is fixed
+      AUTOCLAVED_BASE_URL: 'http://datacollector.infra.ooni.io/ooni-public/autoclaved/' # FIXME: use dataproxy when API moves to AMS
```
hellais: XXX remember to set `REPORT_INDEX_OFFSET` for non beta deployment.

ansible/roles/pguser-readonly/tasks/main.yml
```diff
@@ -0,0 +1,41 @@
+---
+- name: apt install python-psycopg2, to manage postgresql
+  apt:
+    name: python-psycopg2 # IMHO it's OK to have it in dom0 instead of /root/venv, it's rather stable
```
hellais: +1 to adding it to dom0

ansible/roles/ooni-measurements/tasks/main.yml
```diff
@@ -1,19 +1,29 @@
...
+    command: gunicorn --bind 0.0.0.0:{{ oomsm_backend_port }} --workers 4 --timeout 30 measurements.wsgi
+    user: "{{ passwd.oomsmweb.id }}:{{ passwd.oomsmweb.id }}"
```
hellais: XXX we probably also want to do some volume mapping so that we can extract the application logs from outside of the container.

ansible/group_vars/all/vars.yml
```diff
@@ -47,3 +48,4 @@ passwd:
 ansible_venv: /root/venv
 docker_apt_ver: ""
 airflow_tag: "1.8.0-oo3"
+oomsm_tag: "1.0.0-beta.1"
```
hellais: Why did you put the tag in here? I had to resort to grepping the tree in order to find where this was defined.

The convention in other roles is to put this inside of `defaults/main.yml` and if needed override it from the deploy playbook.

ansible/roles/ooni-measurements/tasks/main.yml
darkk: I've not found doc for that.
- [ ] grep flask code

ansible/group_vars/all/vars.yml
darkk: I think, you did not, but 
- [ ] check UID collisions

ansible/group_vars/all/vars.yml
darkk: That's good idea to move it to more reasonable location, I just reused `airflow_tag` location.
- [ ] move tag

ansible/group_vars/all/vars.yml
darkk: Yep, by piwik. Okay, I'll renumber `oomsmweb`.
I suggest following way to take further base UIDs for group of constellation of services to prevent race: `echo $(( 2000 + 50*($RANDOM % 160) ))`.

ansible/roles/ooni-measurements/tasks/main.yml
darkk: @hellais it seems to me, you're mistaken: https://gist.github.com/darkk/f809d7d2a2835ddf97b3a0e15665f412

darkk: On the other hand https://measurements-beta.ooni.io/files/by_date/whalambalamba does not print exception to the webpage so, probably, debug logging is disabled by some of these variables :)

ansible/roles/ooni-measurements/tasks/main.yml
darkk: The application logs it's errors to stderr that goes to `docker logs`. That's probably good enough.

ansible/group_vars/all/vars.yml
darkk: BTW, I'm moving it to role/vars instead of role/defaults as I was unaware of the difference between role/vars and role/defaults and I think that I should review all the stuff I've placed to role/vars here and there :)

PR #126, opened 2017-07-20
Feature/grav

hellais:
This is useful for the `msg.ooni.io` website that users will be redirected to when they receive push notifications.

Depends on #127


ansible/roles/grav/defaults/main.yml
```diff
@@ -0,0 +1,25 @@
+grav_version: "1.3.1"
```
darkk: Depends in piwik

hellais: You mean merging this is dependent on merging piwik?

darkk: Yep.

ansible/roles/grav/defaults/main.yml
```diff
@@ -0,0 +1,25 @@
...
+grav_download_url: "https://github.com/getgrav/grav/releases/download/{{ grav_version}}/grav-admin-v{{ grav_version }}.zip"
+
+# XXX we currently don't create this
+grav_admin_user: grav_admin
+grav_admin_password: "{{ vault_grav_admin_pass }}"
```
darkk: @hellais seems, another vault is missing.

ansible/roles/grav/tasks/main.yml
```diff
@@ -0,0 +1,38 @@
...
+    group: "{{ grav_group }}"
+    state: directory
+  with_items:
+  - "{{ grav_path }}"
+  - "{{ grav_tmp_path }}"
+  - "{{ grav_install_path }}"
+
+- include: setup-php.yml
+- include: setup-nginx.yml
+- include: setup-grav.yml
```
darkk: What's the reason to use some sort of PHP CMS instead of bare static website generator? As far as I see, it's used just to display text to user.

hellais: Because the content of the sites needs to be authored and edited by people that may not want to go through all the hassle of using something like git.
We want to use a CMS so that humans can edit the content.

At some point I would like the OONI website to be refactored to use something like grav.

ansible/roles/grav/defaults/main.yml
darkk: thanks :)

darkk:
I've re-deployed grav, it reported that nginx config was changed:
```
TASK [grav : Configure grav virtual host] **************************************
changed: [msg.ooni.io]
```

Also `msg.ooni.io` says that 

> The Admin plugin has been installed, but no admin accounts could be found

And https://msg.ooni.io/en/messages/2017-07-21-example-page did not survive re-deployment.

darkk:
```
TASK [grav : Create users directory symlink] ***********************************
fatal: [msg.ooni.io]: FAILED! => {"changed": false, "failed": true, "msg": "unsupported parameter for module: dst"}
```

and after s/dst/path/ it becomes
 
```
TASK [grav : Create users directory symlink] ***********************************
fatal: [msg.ooni.io]: FAILED! => {"changed": false, "failed": true, "gid": 2400, "group": "grav", "mode": "0755", "msg": "the directory /srv/grav/settings/user is not empty, refusing to convert it", "owner": "grav", "path": "/srv/grav/settings/user", "size": 95, "state": "directory", "uid": 2400}
```

OK, I'm declaring this playbook broken.

PR #127, opened 2017-07-20
Add role for deploying piwik

hellais:
This is for our analytics server


ansible/roles/piwik/tasks/main.yml
```diff
@@ -0,0 +1,40 @@
...
+  apt: name={{ item }} update_cache=yes state=latest
+  with_items:
+    - 'unzip'
+    - 'vim'
+
+# Environment
+- name: create piwik group
+  group:
+    name: "{{ piwik_group }}"
+    state: present
```
darkk: `uid=2300(piwik) gid=10004(piwik) groups=10004(piwik)`

Also blocked on re-numbering.

ansible/roles/piwik/defaults/main.yml
```diff
@@ -0,0 +1,43 @@
...
+piwik_admin_email: "admin@openobservatory.org"
+
+mysql_host: 127.0.0.1
+mysql_admin_user: root
+mysql_admin_pass: "{{ vault_piwik_mysql_admin_pass }}"
+mysql_piwik_user: piwik
+mysql_piwik_pass: "{{ vault_piwik_mysql_piwik_pass }}"
+mysql_piwik_db: piwik_db
+mysql_piwik_salt: "{{ vault_piwik_mysql_piwik_salt }}"
+mysql_service: mysql
```
darkk: Where is the `vault` file?

ansible/inventory
```diff
@@ -54,6 +54,8 @@ debuilder.infra.ooni.io
 staging.measurements.ooni.io
 demo.probe.ooni.io # NOT-GH (it actually is GH, but not in the shared account) - Note this machine was out of disk space, because older docker containers were not getting deleted. XXX this needs some fixing.
 
+analytics.ooni.io
+
```
darkk: It's wdc, not ams.

ansible/roles/piwik/defaults/main.yml
```diff
@@ -0,0 +1,43 @@
+piwik_download_path: http://builds.piwik.org/piwik.zip
```
darkk: - [ ] Explicit version should be deployed.

ansible/roles/piwik/tasks/setup-piwik.yml
```diff
@@ -0,0 +1,46 @@
+---
+- name: Define release id
+  set_fact:
+    PIWIK_RELEASE_ID: "{{ [ansible_date_time.year, ansible_date_time.month, ansible_date_time.day, ansible_date_time.hour, ansible_date_time.minute, ansible_date_time.second] | join('') }}"
+
+- name: Download files
+  get_url: url={{ piwik_download_path }} dest={{ app_path }}/tmp/piwik.zip
```
darkk: Why is it `zip` and not tar.gz? http://builds.piwik.org/ has tarballs.

hellais: No real reason. I guess I didn't actually check for that URL and just followed the official setup instructions that use this as an example.

ansible/deploy-analytics.yml
```diff
@@ -0,0 +1,26 @@
...
+#        - "{{ passwd.art }}"
+#        - "{{ passwd.darkk }}"
+
+- hosts: analytics.ooni.io
+  gather_facts: true
+  vars:
+    letsencrypt_nginx: yes
+    letsencrypt_domains: "analytics.ooni.io"
+  roles:
+    #- letsencrypt
```
darkk: Why is it disabled?

ansible/roles/piwik/templates/config.ini.php.j2
```diff
@@ -0,0 +1,68 @@
+; <?php exit; ?> DO NOT REMOVE THIS LINE
+; file automatically generated or modified by Piwik; you can manually override the default values in global.ini.php by redefining them in this file.
+[database]
+host = "{{ mysql_host }}"
+username = "{{ mysql_piwik_user }}"
+password = "{{ mysql_piwik_pass }}"
+dbname = "{{ mysql_piwik_db }}"
```
darkk: ansible variables live in global namespace, `mysql_host` sounds like a imperfect name :)

ansible/roles/piwik/templates/php-fpm-www.conf.j2
```diff
@@ -0,0 +1,384 @@
...
+; When not set, the global prefix (or /usr) applies instead.
+; Note: This directive can also be relative to the global prefix.
+; Default Value: none
+;prefix = /path/to/pools/$pool
+
+; Unix user/group of processes
+; Note: The user is mandatory. If the group is not set, the default user's group
+;       will be used.
+user = {{ piwik_user }}
+group = www-data
```
darkk: It's unclear if process should have www-data gid to listen on socket.

ansible/roles/piwik/templates/nginx.conf.j2
```diff
@@ -0,0 +1,84 @@
...
+
+	include /etc/nginx/mime.types;
+	default_type application/octet-stream;
+
+	##
+	# SSL Settings
+	##
+
+	ssl_protocols TLSv1 TLSv1.1 TLSv1.2; # Dropping SSLv3, ref: POODLE
+	ssl_prefer_server_ciphers on;
```
darkk: We should probably define "default" set of protocols / cyphers / DH params and so on for all nginx instances.

ansible/roles/piwik/tasks/setup-piwik.yml
```diff
@@ -0,0 +1,46 @@
...
+    force: yes
+  notify:
+    - restart nginx
+
+- name: Cleanup
+  shell: "ls -dt {{ app_path }}/releases/* | egrep '[0-9]+' | tail -n +6 | xargs rm -rfv"
+  register: deploy_cleanup_result
+  changed_when: "'removed' in deploy_cleanup_result.stdout"
+  sudo: yes
+  sudo_user: root
```
darkk: `/srv/piwik/releases/*/piwik` files have some mix of piwik and www-data permissions.
I don't understand the reason for these files to be owned by `piwiki` (I don't see explicit mention for that)
Also, it looks like bug to me -- php website should not be able to modify it's scripts, that's common "infection" technique.

ansible/roles/piwik/defaults/main.yml
```diff
@@ -0,0 +1,43 @@
...
+
+piwik_admin_user: piwik_admin
+piwik_admin_password: "{{ vault_piwik_admin_pass }}"
```
darkk: I've not found any uses of `piwik_admin_password` variable.

ansible/roles/piwik/templates/nginx-site-piwik.j2
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  root  {{ app_path }}/current/piwik;
+  index  index.php index.html;
+  # Disallow any usage of piwik assets if referer is non valid.
+  location ~* ^.+\.(?:jpg|png|css|gif|jpeg|js|swf)$ {
+    # Defining the valid referers.
+    #valid_referers none blocked *.mysite.com othersite.com;
+
+    expires max;
```
darkk: It really looks like bug. It means that `//analytics.ooni.io/piwik.js` NEVER changes and is cached forever.
Does it sound like true statement?

darkk:
I've **NOT** re-deployed piwik as current setup is fragile and seems to have fatal bugs (e.g. `expire max`), seems, it should be eventually cleaned up and setup on a different domain name :-/

ansible/roles/piwik/tasks/main.yml
darkk: Done

ansible/roles/piwik/defaults/main.yml
darkk: Fixed.

ansible/inventory
darkk: Fixed & enforced.

ansible/roles/piwik/templates/nginx-site-piwik.j2
hellais: I don't think this is an issue, because we don't use the image based tracking of piwik, but only the js based one.

Since E-Tag is set and the piwik JS script is fetched via javascript the caching rule should be good. If for some reason there is the need to update the software I believe E-Tag should be updated and the browser cache will be busted. If we need to force busting of the cache we can always do something like `/piwik.js?$RANDOM_NONCE`.

hellais: Maybe E-Tag may actually not be enough. However I don't see this as being fatal, it just means that if we make an update to piwik the client side code will not be updated in browser that have piwik.js already cached. I don't think that is the end of the world to be honest as I would expect the client code to be backward compatible. In any case we can deploy cache breaking solution (as outlined above) without having to redeploy on another domain if we care about this.

darkk: > and the browser cache will be busted

As far as I know, `Expires: Thu, 31 Dec 2037 23:55:55 GMT` means that browser does not have to check ETag till 2037.

darkk: > I don't see this as being fatal

True. I've not thought about nonce hack for some reason.

> I would expect the client code to be backward compatible

I would also expect any software to have subtle bugs in compatibility areas :-P

PR #129, opened 2017-07-29
fix,refactor(proteus-events): topics correctness

bassosimone:
1) As stated in TheTorProject/proteus#13, we must not set a topic
for Android because that leads to FCM rejecting the message.

2) Define two different topics for iOS, one for production and
one for testing, because I guess we'll need them soon.


darkk:
LGTM. Further comments will be at `feature/orchestration` branch :)

PR #130, opened 2017-07-31
feature(roles/adm): ssh -= {pwd-based-auth, root-login}

bassosimone:
This diff disabeles password-based-authentication for SSH. as
well as the possibility of logging in directly as root.

While there, add to .gitignore annoying files on macOS as
well as the location where I saved the virtualevn (in many
cases I've seen virtualevens called like this, so I'd argue
this is not surprising at all).

Based on: https://ryaneschinger.com/blog/securing-a-server-with-ansible/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.DS_Store
 logging.cfg
 *.log
 invoke.yaml
@@ -6,3 +7,4 @@ invoke.yaml
 *.swo
 backend-files/
 secrets.env.j2
+/venv
--- a/ansible/roles/adm/handlers/main.yml
+++ b/ansible/roles/adm/handlers/main.yml
@@ -1,0 +1,3 @@
+---
+- name: Restart ssh
+  service: name=ssh state=restarted
--- a/ansible/roles/adm/tasks/main.yml
+++ b/ansible/roles/adm/tasks/main.yml
@@ -39,3 +39,17 @@
 
 - name: sudoers.d/adm
   template: src=sudoers dest=/etc/sudoers.d/adm owner=root group=root mode=0440 validate='visudo -cf %s'
+
+- name: lockdown access to ssh
+  lineinfile: dest=/etc/ssh/sshd_config
+              regexp="^PasswordAuthentication"
+              line="PasswordAuthentication no"
+              state=present
+  notify: restart ssh
+
+- name: do not allow root to login via ssh
+  lineinfile: dest=/etc/ssh/sshd_config
+              regexp="^PermitRootLogin"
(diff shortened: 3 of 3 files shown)


.gitignore
```diff
@@ -1,8 +1,10 @@
...
 *.log
 invoke.yaml
 *.pyc
 *.swp
 *.swo
 backend-files/
-secrets.env.j2
\ No newline at end of file
+secrets.env.j2
+/venv
```
darkk: Where is /venv from? Is it your local `venv` for specific ansible version?

ansible/roles/adm/tasks/main.yml
```diff
@@ -39,3 +39,17 @@
...
+              line="PasswordAuthentication no"
+              state=present
+  notify: restart ssh
+
+- name: do not allow root to login via ssh
+  lineinfile: dest=/etc/ssh/sshd_config
+              regexp="^PermitRootLogin"
+              line="PermitRootLogin no"
+              state=present
+  notify: restart ssh
```
darkk: IMHO, that's bad idea to drop `root` login in `adm` role as it's played while logging in as root :)
I suggest to keep `adm` role doing nothing but creation of administrator's accounts and move various defaults to separate role.

.gitignore
bassosimone: Yep.

ansible/roles/adm/tasks/main.yml
bassosimone: Okay. I should I call the role? `hardening`?

darkk: IDK. Maybe `dom0` as this role will accumulate all "ooniized" defaults for host system, including mentioned at #101.

bassosimone: Okay, I will give this a spin later. Thanks!

darkk:
`PermitRootLogin no` can't be used right now as we have lots of keys in /root/.ssh/authorized_keys{,2}, some of these keys are actually used and, what's even worse, some of these keys are used for various periodic jobs.

I suggest following:
- `LogLevel VERBOSE` to log key fingerprints to auth.log
- `PermitRootLogin without-password`

ansible/roles/adm/handlers/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+- name: Restart ssh
```
darkk: BTW, I've just noticed that `Restart ssh` != `restart ssh` :)

darkk:
Closed in favor of #139

PR #136, opened 2017-08-11
Prometheus agent

darkk:
That's already deployed to `all:!no_nodeexp`, but feel free to comment if you want to :)


darkk:
@bassosimone @hellais @anadahz I'm merging that as it's already deployed, but your comments are still welcome (as usual) if you see something interesting :)

ansible/roles/dom0-defaults/tasks/main.yml
```diff
@@ -0,0 +1,18 @@
...
+- name: install some common packages
+  apt:
+    name: "{{ item }}"
+    state: latest
+    update_cache: yes
+    cache_valid_time: 28800
+    install_recommends: false
+  with_items:
+    - ca-certificates
+    - tmux
```
hellais: Can we haz screen too?

darkk: Sure, but why? It's ancient software that is often installed as suid binary and that still can't handle UTF-8 in sane way sometimes.
I'm not against that (and nothing can prevent you from installing it :-P), but I would like to know what use-case is covered better with `screen`.

hellais:
Since this was merged I am having issues using the vault command:

what I get is the following:

```
./vault edit host_vars/prometheus.infra.ooni.io/vault
++ git rev-parse --show-toplevel
+ gitroot=/Users/x/code/OONI/ooni-sysadmin
+ ansible-vault --vault-password-file /Users/x/code/OONI/ooni-sysadmin/ansible/password-pipe edit host_vars/prometheus.infra.ooni.io/vault
gpg: WARNING: server 'gpg-agent' is older than us (2.0.30 < 2.1.21)
+ '[' edit '!=' view ']'
+ '[' edit '!=' decrypt ']'
+ '[' edit '!=' rekey ']'
+ /Users/x/code/OONI/ooni-sysadmin/scripts/mk-travis-vault
++ git rev-parse --show-toplevel
+ gitroot=/Users/x/code/OONI/ooni-sysadmin
+ ansible-vault --vault-password-file /Users/x/code/OONI/ooni-sysadmin/ansible/password-pipe view /Users/x/code/OONI/ooni-sysadmin/ansible/.travis.vaultpw
ssh gpg: WARNING: server 'gpg-agent' is older than us (2.0.30 < 2.1.21)
a+ '[' view '!=' view ']'
+ set -o errexit
+ cd /Users/x/code/OONI/ooni-sysadmin/ansible
++ find . -type f -name vault '!' -perm /111
++ sed 's,/vault$,/.travis.vault,'
find: -perm: /111: illegal mode string
+ make -j1 -f /dev/stdin
make: *** No targets.  Stop.
```

I suspect it has something to do with the fact that BSD `find` is different from linux `find`.

darkk:
That's true: 1e64b4133f1cf35f3b5fd27ee60dac34d56cafd5

ansible/roles/dom0-defaults/tasks/main.yml
hellais: > I'm not against that (and nothing can prevent you from installing it :-P), but I would like to know what use-case is covered better with screen.

Nothing particular, it's just a matter of habit. I know how to use screen fairly well, but know how to use tmux far less.

darkk: It's mostly the same with ^B default instead of ^A (so readline's shortcut go-to-beginning-of-the-line works) and some nice features like tiled layout of split-screen and sending keyboard input to every pane (that's nice to feature to run same interactive command on 9..12 servers).
I've added screen, but I'd suggest to give `tmux` a try :)

PR #146, opened 2017-08-21
test env for ntp test & public ext/known_hosts

darkk:


darkk:
@bassosimone @anadahz I added ext/known_hosts that is auto-updated with ssh, so it can be added to ~/.ssh/config

PR #154, opened 2017-08-30
Feature/jupyter

hellais:


hellais:
@darkk Is it really the case that the inventory has a harcoded hash:

```
    "assertion": "groups.all | sort | join('|') | hash('sha256') == '2f901f46fcf5b1c09757db99e088d9c62a1148dfe70c0ad044a7100e5631fcdf'", 
```
How do I update it?

darkk:
> How do I update it?

You verify that `./play inventory-check.yml` passes the check for all hosts in the inventory, and if it does, you update the hash.

It's there as an annoying protection against:
- hosts in inventory that don't have `dom0-bootstrap.yml` playbook applied to them
- dead hosts in inventory
- hosts placed to wrong DC in inventory

ansible/roles/jupyter/tasks/setup-anaconda.yml
```diff
@@ -0,0 +1,67 @@
...
+
+- name: install pkgs with pip
+  become: yes
+  become_user: root
+  command: '{{anaconda_link_dir}}/bin/pip install {{ item }}'
+  with_items: "{{ pip_pkgs }}"
+
+- name: remove conda-curl since it conflicts with the system curl
+  become: yes
+  become_user: root
```
darkk: FYI: `become` and `become_user` are likely overridden by global variable `ansible_become: true`, that's mentioned in 1f8a9681331d07c95ab7ae3fd995406341068326 but I've lost link to corresponding ansible

ansible/roles/jupyter/templates/jupyter_notebook_config.py.j2
```diff
@@ -0,0 +1,605 @@
...
+#c.NotebookApp.open_browser = True
+
+## Hashed password to use for web authentication.
+#
+#  To generate, type in a python/IPython shell:
+#
+#    from notebook.auth import passwd; passwd()
+#
+#  The string should be of the form type:salt:hashed-password.
+c.NotebookApp.password = "{{ jupyter_password_hash }}"
```
darkk: :+1:

hellais:
I am going to merge this as it's deployed and travis is passing.

darkk:
FYI, prometheus configuration was not updated, so `jupiter.ooni.io` metrics were not scraped.
I've done that running `./play deploy-prometheus.yml -t prometheus-conf`.

hellais:
For my future self:

We may eventually want to use jupyterhub which seems like the thing you want for multi-user setups: http://jupyterhub.readthedocs.io/en/latest/

PR #161, opened 2017-09-12
Use explicit list of alert recipients instead of team@ or admin@

darkk:
We have no transparent rule to configure email aliases, so it's better to be safe and use flattened list instead of admin@ in systems allowing explicit list of addresses.

I don't think that x4 increase of Amazon SES bill matters. That's still dozens of emails, not thousands.

CC: @bassosimone, @hellais, @anadahz


hellais:
LGTM.

PR #167, opened 2017-09-27
Slacktopus update

hellais:
Add #ooni-dev, remove secret channels from mapping.


ansible/host_vars/slack.openobservatory.org/vars.yml
```diff
@@ -45,10 +44,8 @@ slack_irc_config: |
       "server": "irc.oftc.net",
       "token": "{{ slack_irc_token }}",
       "channelMapping": {
-        {% for channel in secret_channels %}
-        "{{ channel.slack }}": "{{ channel.irc }}",
-        {% endfor %}
         "#ooni": "#ooni",
+        "#ooni-dev": "#ooni-dev",
```
darkk: I'm unsure if #ooni-bots should be bridged. I'd rather drop that channel as current code of the bridge does not relay Alertmanager messages to IRC at all.

hellais: I think that is reasonable. It's mostly useful to us and we are using the IRC bridge anyways for other channels.

Less channels we have to maintain with broken support the better I think.

ansible/host_vars/slack.openobservatory.org/vars.yml
```diff
@@ -45,11 +44,8 @@ slack_irc_config: |
       "server": "irc.oftc.net",
       "token": "{{ slack_irc_token }}",
       "channelMapping": {
-        {% for channel in secret_channels %}
-        "{{ channel.slack }}": "{{ channel.irc }}",
-        {% endfor %}
```
darkk: Seems, it was already done in https://github.com/TheTorProject/ooni-sysadmin/commit/7d5bbf25349b5c9c904fe0740ec4ab7848365072 that is already in `master`...
Poor `ooni-sysadmin` repo, it's full of weeds... :cry:

darkk:
I have no idea if this PR is useful or not. I'll just close it, change list of channels to reflect some list that seems "sane" to me and that'll be it.

For the record, `#ooni-dev` was not bridged since 2017-10-24 08:46:33 UTC.

PR #169, opened 2017-09-28
Deploy ooni-api 1.0.0

hellais:


ansible/deploy-measurements.yml
```diff
@@ -20,11 +20,11 @@
       name: oomsm-beta
```
darkk: `letsencrypt_domains` is not updated.

Actually, changing hostname in inventory and hosts does not matter that much, but chaning `letsencrypt_domains` matters as it's used for SSL cert.

hellais: Done

hellais:
In the conflict resolution I think the hash needs to be updated.

PR #181, opened 2017-10-25
Migrate to the new s3 bucket

hellais:
This depends on: https://github.com/TheTorProject/ooni-pipeline/pull/101


ansible/roles/airflow/files/docker-trampoline
```diff
@@ -92,12 +93,20 @@ case "$script" in
...
     ;;
 
+    sanitised_s3_cp)
+    docknet="" # no `--network none'
+    volargs="${volargs} --env-file=/etc/af-worker/s3read.env"
+    volargs="${volargs} --volume=$public/sanitised/${bucket}:$public/sanitised/${bucket}:ro"
+    script=""
+    set -- /usr/local/bin/aws_s3_cp.py --src "$public/sanitised/${bucket}" --dst "s3://ooni-data/sanitised/${bucket}"
+    ;;
+
```
darkk: You're either kidding me or have not looked at the data and/or misunderstand how it works (every option is OK, but I truly don't understand what one of them is the case).
It obviously does not work like that, the `.../sanitized/2017-10-*` directories are empty.

hellais: 1. Chillax
2. Most directories on datacollector are in fact empty, but I was assuming this was due to some bug in the new pipeline as I do see some of them containing files. Example:
```
$ ls  -1 /data/ooni/public/sanitised/2017-02-24 | wc -l
2470
```
3. If that is not the correct place to get these from, please update the path to what is the correct path to get them from.

darkk:
- [ ] revoke previous key when ready

darkk:
Another thing that makes me a bit unhappy is `s3read_access_key` that has `AmazonS3FullAccess` capability.

hellais:
> Another thing that makes me a bit unhappy is s3read_access_key that has AmazonS3FullAccess capability

The s3 access you key are talking about, is the one for the `ooni-pipeline` user. I didn't bother to change the name of the `env` file (or the environment variable), but it should be `rw` capability enabled key, otherwise you can't write to the s3 bucket.

ansible/host_vars/datacollector.infra.ooni.io/vars.yml
```diff
@@ -10,10 +10,14 @@ airflow_fernet_key: "{{ vault_airflow_fernet_key }}"
 airflow_redis_password: "{{ vault_airflow_redis_password }}"
 airflow_postgres_password: "{{ vault_airflow_postgres_password }}"
 
-# tokens to read https://s3.amazonaws.com/ooni-* buckets
+# tokens to read https://s3.amazonaws.com/ooni-* buckets from Open Observatory CEO personal S3 account
 s3read_access_key_id: "{{ vault_s3read_access_key_id }}"
 s3read_secret_access_key: "{{ vault_s3read_secret_access_key }}"
```
darkk: @hellais is that reference good enough? I used the one that was used in the mail thread with Jim Ying :-)

ansible/roles/airflow/files/docker-trampoline
```diff
@@ -97,7 +98,7 @@ case "$script" in
     volargs="${volargs} --env-file=/etc/af-worker/s3read.env"
     volargs="${volargs} --volume=$public/sanitised-s3-ls:$public/sanitised-s3-ls:rw"
     script=""
-    set -- /usr/local/bin/aws_s3_ls.py --url "s3://ooni-public/sanitised/$bucket" --s3-ls "$public/sanitised-s3-ls/${bucket}.json.gz"
+    set -- /usr/local/bin/aws_s3_ls.py --url "s3://ooni-data/sanitised/$bucket" --s3-ls "$public/sanitised-s3-ls/${bucket}.json.gz"
```
darkk: NB: publishing of `sanitised` files is not fixed yet. Maybe it never will, something is triggering the bug in luigid that prevents that pipeline from progressing on 2017-10-09 for 36h+ I don't feel like that the investigation is well-spent time.

darkk:
Merged as it's deployed.

The way to backfill was `airflow backfill --ignore_dependencies --start_date 2012-12-05 --end_date 2017-11-08 -t autoclaved_.*s3_sync --pool datacollector_disk_io hist_canning` as I expected these actions to be not that CPU-intensive, but...
- actually `aws s3 sync` takes fair amount of CPU
- as already known, `--pool` is not respected

darkk:
> The way to backfill was `airflow backfill`

And then it renamed DagRun's to `backfill_DADADATE`. And then the name is used as a [special string](https://github.com/apache/incubator-airflow/blob/1.8.0+apache.incubating/airflow/models.py#L4157-L4162). And that special string has [special-case handling](https://github.com/apache/incubator-airflow/blob/1.8.0+apache.incubating/airflow/jobs.py#L1057-L1060) in the scheduler.

:disappointed:

PR #185, opened 2017-11-20
explorer: A/B test against ooni-measurements API compatibility layer

darkk:


scripts/docker-build
```diff
@@ -0,0 +1,61 @@
...
+fi
+
+# git@github.com:TheTorProject/ooni-explorer.git -> explorer
+# git@github.com:TheTorProject/ooni-pipeline.git af/shovel/ -> pipeline-shovel
+if [ -n "$prefix" ]; then
+    img_slug="${repo}-${prefix}"
+else
+    img_slug="${repo}"
+fi
+img="openobservatory/$img_slug"
```
darkk: I don't want to discuss naming.
That's "private" images, so I just use repo name. IMHO exact name does not matter, but having uniform names is something good.

scripts/docker-build
```diff
@@ -0,0 +1,61 @@
...
+
+ver="$1"
+if [ -z "$ver" ]; then
+    ver="$(date '+%Y%m%d')"
+fi
+ver="${ver}-${git_head8}" # tags in git repo are not used, so HEAD is stored as image version
```
darkk: Also, I don't want to discuss versioning.
That's "private" images, so I just use date + git hash and ensure that it's really GIT HASH that is used to built the image (and not random working tree state).

PR #186, opened 2017-11-27
Update/ooni api

hellais:
In here I do two things:

* Bump up the proxy timeout for nginx to 3 minutes (yeah it's super long, but it's better to do this than have a bunch of 500 errors)
* Try to add new domains to letsencrypt chain

### DIFF
--- a/ansible/deploy-measurements.yml
+++ b/ansible/deploy-measurements.yml
@@ -8,7 +8,7 @@
       letsencrypt_nginx: true
       letsencrypt_pre_hook: ''
       letsencrypt_post_hook: '' # FIXME: I don't want to spend time understanding the role :-(
-      letsencrypt_domains: api.ooni.io
+      letsencrypt_domains: "api.ooni.io,measurements.ooni.io,measurements.ooni.torproject.org"
 
 - hosts: hkgmetadb.infra.ooni.io
   gather_facts: false # no useful facts there :)
--- a/ansible/inventory-check.yml
+++ b/ansible/inventory-check.yml
@@ -99,6 +99,5 @@
       run_once: true
 
     - name: stamp the inventory that was checked
-      assert: { that: groups.all | sort | join('|') | hash('sha256') == '4debade14b752f96878db0b3692eef6a2553fb2d5ed0c6580daa00cc03fa004f' }
+      assert: { that: groups.all | sort | join('|') | hash('sha256') == '39a44c1d9937fb252d2323e5b777467e430095697344429767103abd6f44637b' }
       run_once: true
-...
--- a/ansible/roles/ooni-measurements/tasks/main.yml
+++ b/ansible/roles/ooni-measurements/tasks/main.yml
@@ -22,6 +22,6 @@
       APP_ENV: production
       DATABASE_URL: 'postgresql://oomsm-beta:{{ oomsm_beta_postgres_password }}@hkgmetadb.infra.ooni.io/metadb' # FIXME: ?sslmode=require
       AUTOCLAVED_BASE_URL: 'http://datacollector.infra.ooni.io/ooni-public/autoclaved/' # FIXME: use dataproxy when API moves to AMS
-    command: gunicorn --bind 0.0.0.0:{{ oomsm_backend_port }} --workers 5 --timeout 60 measurements.wsgi
+    command: gunicorn --bind 0.0.0.0:{{ oomsm_backend_port }} --workers 5 --timeout {{ oomsm_request_timeout }} measurements.wsgi
     user: "{{ passwd.oomsmweb.id }}:{{ passwd.oomsmweb.id }}"
     restart_policy: unless-stopped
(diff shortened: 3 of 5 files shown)


ansible/deploy-measurements.yml
```diff
@@ -8,7 +8,7 @@
       letsencrypt_nginx: true
       letsencrypt_pre_hook: ''
       letsencrypt_post_hook: '' # FIXME: I don't want to spend time understanding the role :-(
-      letsencrypt_domains: api.ooni.io
+      letsencrypt_domains: "api.ooni.io,measurements.ooni.io,measurements.ooni.torproject.org"
```
darkk: As far as I remember, we agreed that
- moving api.ooni.io to VM taken by measurements.ooni.io is bad idea
- redirecting measurements.ooni.io to api.ooni.io should be done after access_log review
Was access_log reviewed?

hellais: > moving api.ooni.io to VM taken by measurements.ooni.io is bad idea

Why is this the case?

> redirecting measurements.ooni.io to api.ooni.io should be done after access_log review

Fun fact: we don't actually have any access_logs. Though even if we did, what would be looking for in them exactly?

darkk: > Why is this the case?

We've discussed that @ Slack on the 27th of November:

> 15:42 darkk> also, if you want to move ooni-api to ams to reduce latency you'll unlikely achieve your goal till DB replica is also in AMS and it's blocked by explorer that is in queue right after TorBro.
> 15:43 hellais> At least it's going to be slightly better for non DB backed responses
> 15:44 hellais> But that's a fair point
> 15:44 darkk> oh.... I'm not so sure regarding other responses

> we don't actually have any access_logs

We do. They're just a bit unusually formatted and placed in a bit unusual place :-D
`/var/lib/docker/containers/ffee8022e2383b085fb0d27836ce063f406964075227c543d6c7daa186e31673/ffee8022e2383b085fb0d27836ce063f406964075227c543d6c7daa186e31673-json.log`

> what would be looking for

Are there any API consumers? I mean some input to make a decision if we should redirect (where possible) or just place `410 Gone` with link to `api.ooni.io`. E.g. `/files/download/xxxx` is not trivially redirectable as it has no bucket name.

hellais: > Do you remember that discussion?

Yes, but if you look at the last message I wrote, I said "At least it's going to be slightly better for non DB backed responses". As in I get your point, but I still think it's going to be a minor improvement and we should go ahead and do it.

Sorry if I wasn't clear.

hellais: > E.g. /files/download/xxxx is not trivially redirectable as it has no bucket name

Actually this exists in the new API as well and just serving it from `api.ooni.io` will work: https://github.com/TheTorProject/ooni-measurements/blob/master/measurements/pages/__init__.py#L269.

I believe as a matter of fact most things should just work out of the box if we just point the A record for measurements.ooni.io to api.ooni.io or do a redirect of measurements.ooni.io/path to api.ooni.io/path.

darkk: > Actually this exists

Yep. It exists, but it does not work "out of box" with plain redirect rewriting domain or prefix.
Also, old URLs are also kinda "gone".
That's why I say, that this activity needs some time to grep logs and check if some redirections we're going to put there actually work as expected, that's why I say that it's a bit more complex than swapping A records if the goal is to preserve URLs :-)

```
$ curl --head https://measurements.ooni.io/files/download/20170313T111025Z-DK-AS9158-http_invalid_request_line-20170313T111027Z_AS9158_vp1nSAqvabxrzrdyRUPPjruyBgSrZMXMmcyfMvAn3JEeyw8kSD-0.2.0-probe.json
HTTP/1.1 302 FOUND
Location: https://s3.amazonaws.com/ooni-public/sanitised/2017-03-14/20170313T111025Z-DK-AS9158-http_invalid_request_line-20170313T111027Z_AS9158_vp1nSAqvabxrzrdyRUPPjruyBgSrZMXMmcyfMvAn3JEeyw8kSD-0.2.0-probe.json

$ curl --head https://s3.amazonaws.com/ooni-public/sanitised/2017-03-14/20170313T111025Z-DK-AS9158-http_invalid_request_line-20170313T111027Z_AS9158_vp1nSAqvabxrzrdyRUPPjruyBgSrZMXMmcyfMvAn3JEeyw8kSD-0.2.0-probe.json
HTTP/1.1 404 Not Found

$ curl --head https://api.ooni.io/files/download/20170313T111025Z-DK-AS9158-http_invalid_request_line-20170313T111027Z_AS9158_vp1nSAqvabxrzrdyRUPPjruyBgSrZMXMmcyfMvAn3JEeyw8kSD-0.2.0-probe.json
HTTP/1.1 404 NOT FOUND

$ curl --head https://api.ooni.io/files/download/2017-03-14/20170313T111025Z-DK-AS9158-http_invalid_request_line-20170313T111027Z_AS9158_vp1nSAqvabxrzrdyRUPPjruyBgSrZMXMmcyfMvAn3JEeyw8kSD-0.2.0-probe.json
HTTP/1.1 200 OK
```

hellais: > Also, old URLs are also kinda "gone".

Ok that is definitely a bug in the API. I will look into fixing it. See: https://github.com/TheTorProject/ooni-measurements/issues/44

hellais:
Discussed this with @darkk it can be closed and the branch deleted.

PR #198, opened 2018-01-16
Update/proteus

hellais:
This updates proteus-orchestrate to `v0.2.0-rc.1`.

It also renames the proteus-events role to proteus-orchestrate and changes the path of the downloaded .tar.gz to match up with the new release process.


bassosimone:
@hellais I think you should probably update some inventory to unbreak travis.

ansible/inventory
```diff
@@ -57,6 +57,7 @@ msg.ooni.io
 notify.proteus.ooni.io
 events.proteus.ooni.io
 registry.proteus.ooni.io
+orchestrate.ooni.io
```
bassosimone: Sanity: is this host in AMS?

ansible/inventory
```diff
@@ -97,6 +98,7 @@ db-1.proteus.test.ooni.io
 proteus.test.ooni.io
 notify.proteus.ooni.io
 events.proteus.ooni.io
+orchestrate.ooni.io
```
bassosimone: I assume `psk_orchestration` means pre-shared-key orchestration? It seems to group all orchestration hosts?

ansible/inventory
hellais: Yes. It's actually the same host as events.proteus.ooni.io. After I confirm it deploys cleanly I will remove events.proteus.ooni.io.

ansible/roles/proteus-orchestrate/defaults/main.yml
```diff
@@ -11,16 +11,16 @@ proteus_environment: "production"
...
 proteus_notify_click_action_android: "org.openobservatory.ooniprobe.OPEN_BROWSER"
 
-proteus_version: "0.1.0-beta.9"
+proteus_version: "0.2.0-rc.1"
 proteus_download_url: https://github.com/thetorproject/proteus/releases/download
 
-proteus_tarball_url: "{{ proteus_download_url }}/v{{ proteus_version }}/proteus-events-{{ proteus_version }}.linux-{{ go_arch }}.tar.gz"
-proteus_path: "{{ proteus_install_path }}/proteus-events-{{ proteus_version }}.linux-{{ go_arch }}"
+proteus_tarball_url: "{{ proteus_download_url }}/v{{ proteus_version }}/proteus_{{ proteus_version }}_linux_{{ go_arch }}.tar.gz"
+proteus_path: "{{ proteus_install_path }}/proteus_{{ proteus_version }}_linux_{{ go_arch }}"
```
bassosimone: I assume this is a recent filenames normalization I remember I did read about

ansible/inventory
hellais: That's what the comment says :P

ansible/roles/proteus-orchestrate/templates/proteus-orchestrate.toml.j2
```diff
@@ -13,13 +13,7 @@ gorush-basic-auth-password = "{{ proteus_notify_basic_auth_password }}"
...
 
 [database]
 url = "{{ proteus_database_url }}"
-active-probes-table = "active_probes"
-probe-updates-table = "probe_updates"
-jobs-table = "jobs"
-job-alerts-table = "job_alerts"
-job-tasks-table = "job_tasks"
-tasks-table = "tasks"
-accounts-table = "accounts"
```
bassosimone: I remember reading about this refactoring.

ansible/roles/proteus-orchestrate/templates/proteus-orchestrate.toml.j2
```diff
@@ -13,13 +13,7 @@ gorush-basic-auth-password = "{{ proteus_notify_basic_auth_password }}"
 [api]
 port = {{ proteus_api_port }}
 address = "{{ proteus_api_address }}"
+default-inputs-to-return = 10
```
bassosimone: I don't remember reading about this key, but I assume it's implemented.

ansible/inventory
bassosimone: ah, it was very above in the diff. I expanded until `psk_orchestration` but no upper than that.

ansible/roles/proteus-orchestrate/templates/proteus-orchestrate.toml.j2
hellais: I should actually probably drop this from the default config and here, as I removed from the source since I don't think it makes sense to have config defined defaults (if it's a default value for something it should not be overridable from the config...)

bassosimone: > I don't think it makes sense to have config defined defaults

Sounds very correct to me. Please zap that!

ansible/inventory
darkk: > Sanity: is this host in AMS?

FYI `inventory-check.yml` has that check.

> It's actually the same host as events.proteus.ooni.io

Please, don't duplicate hosts in inventory.

> After I confirm it deploys cleanly I will remove events.proteus.ooni.io.

I suggest to avoid renames in `inventory` when possible, it leads to mistakes.

PR #199, opened 2018-01-16
new ansible, postgres config and some fixes

darkk:
FYI, that's already deployed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,7 +236,24 @@ ooniprobe --collector httpo://CollectorAddress.onion blocking/http_requests \
 
 # When adding new hosts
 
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -256,6 +273,8 @@ Be also to then run:
 
 This will fetch updates to the DNS zone.
 
+`inventory_hostname` MUST NOT be renamed.
+
```
darkk: @hellais @bassosimone do you confirm, that we've agreed on that naming policy three and a half months ago? :)

bassosimone: I am not sure. What do you mean?

PR #223, opened 2018-07-31
Add script for generating OONI compatible hostnames

hellais:
While waiting for things to deploy...


scripts/wordlists/slavic-deity-names.txt
```diff
@@ -0,0 +1,97 @@
+Rod
```
SuperQ: It'd make sense to `sort` these.

hellais: I left them in the order in which they appear on wikipedia. Since we pick from the list in random order, it doesn't affect how they are chosen, but it maybe makes it easier to check if some other deities are added to the wikipedia page in the future.

hellais: Whatever I am just going to sort them. I doubt I will ever go back to this list to update it :P

PR #231, opened 2018-10-19
Add @lorenzoPrimi to events.proteus.test.ooni.io

hellais:


darkk:
Hm... mtime suggests that the account was already there for some time:
```
-r-------- 1 nuke nuke 425 Aug  4  2017 /home/nuke/.ssh/authorized_keys
```

hellais:
> Hm... mtime suggests that the account was already there for some time:

Maybe on notify, but I don't think it was on events.

darkk:
> Maybe on notify

True. I was mistaken.

ansible/host_vars/events.proteus.test.ooni.io/vars.yml
```diff
@@ -0,0 +1,2 @@
+---
+adm_logins: [ art, darkk, sbs, nuke ]
```
darkk: There is no `weasel` login here. Does it mean that it should be dropped from [`all`](https://github.com/ooni/sysadmin/blob/master/ansible/group_vars/all/vars.yml#L105) as well?

PR #232, opened 2018-10-19
setting production false and topic fix

lorenzoPrimi:
I have a new dev certificate, contact me to get it


hellais:
@lorenzoPrimi do we still need this?

hellais:
If so please ping me on slack about it and I will get you sorted.

PR #234, opened 2018-10-22
Create PostgreSQL Server CA, add sync-test-lists to airflow

darkk:


ansible/inventory
```diff
@@ -138,6 +143,8 @@ hkgmetadb.infra.ooni.io
 datacollector.infra.ooni.io
 hkgmetadb.infra.ooni.io
 amsmetadb.ooni.nu
+# db-1.proteus.test.ooni.io
+# db-1.proteus.ooni.io
```
darkk: I added `log_connections` so I plan to roll out firewall on these hosts tomorrow.
@hellais tell me, if that's a bad idea.

ansible/inventory
```diff
@@ -127,6 +127,11 @@ registry.proteus.ooni.io
 db-1.proteus.ooni.io
 proteus.ooni.io
```
darkk: I tried to clean up all the tokens shared across all these 10 hosts of this group, but failed to do that _quickly_.
So, I just added another FIXME...

Issue #226, opened 2018-09-05
Monitoring epic, Oct 2017 … Sep 2018


darkk:
This ticket tracks enhancements to OONI infra monitoring through Oct 2017 … Sep 2018.

- [x] replace basic authentication in Prometheus with client SSL cert validation -- 23883a2104c1733b56c036c87ffc82754f60c6a3 & #155
- [x] configure cross-monitoring with another Prometheus -- 3bf09a7298666756ff73a213ad6d56658ed796e1
- [x] check that all inventory hosts are monitored -- d226bd144e0600cb441d1acdfb8ca2e0e91181e4
- [x] improve disk space alerts for filesystems over 50GiB -- 7c57b5bb16520bd6788bed1dc5af1731f6120042
- [x] monitor rare periodic jobs with push-gateway (e.g. `hellais-nas` backup) -- d6ba8fc4ef809fb49aace3308c1ae9ec5f0082c1
- [x] monitor IO stalls at GH, iowait% -- 085fd5a8e75018931c410eceed8f8b5171f0dec2
- [x] monitor IO stalls at GH, seeksample -- e342b1d50a7afb0857f93e057da811d2aff883f9
- [x] define "canonical" TSV log format for nginx -- edae501943e2c80fbd30ee4aaab3e5627a7b7ff0
- [x] monitor node_exporter/textfile liveness -- 7eff51c5781a0f4218560e6c008331bf47c6e19c
- [x] monitor nginx upstream request latency -- 6e65c46af2ec3f742e87b167566ee6fc5381d723
- [x] monitor airflow task failures -- 3f7cfac024bc8accbfcdcd314e8b952742215e75
- [x] speedup DiskWillFillInAWeek rule -- fd6318151df93df56a9e0553438113f99f6554c4
- [x] check that all GH host are stored in inventory -- 8a12b5ae9e4bd938e55653fed8911b9d8eb2be08
- [x] improve DiskWillFillInAWeek alerts -- 01daac85353440d61eb81b7494be0b8996e7e55b, b1b3995dcb435abfefb4922e3a659244c59abe83
- [x] monitor if OONI Website mirrors diverge -- f4841ef86af97cc4142275b41481cbc780cf3f45
- [x] deliver email alerts to MX (without smarthost) -- 2d220b23dcb7c7e07abd3e6103a8922a52d2fed4 #159
[...]

PR #238, opened 2018-10-31
More monitoring & alerting from 2018-epic

darkk:
That's another a PR for an umbrella #226


darkk:
All of that is deployed.

PR #242, opened 2018-11-12
Feature/explorer next

hellais:


scripts/docker-build
```diff
@@ -34,7 +34,12 @@ if [ -n "$prefix" ]; then
 else
     img_slug="${repo}"
 fi
-img="openobservatory/$img_slug"
+
+if [ -z ${IMG_SLUG} ]; then
+    img="openobservatory/$img_slug"
+else
+    img="openobservatory/${IMG_SLUG}"
+fi
```
darkk: Why do you need to override that?
The reason to codify that was to have explicit link between repository name and docker image name for readability, maintainability, grepability and all  other alike *abilities.

ansible/roles/ooni-explorer-next/defaults/main.yml
```diff
@@ -0,0 +1,3 @@
+explorer_tag: '20181112-238cac95'
+explorer_backend_ipv4: 172.28.14.254
+explorer_backend_port: 3000
```
darkk: It's not a big deal, but BCP is to avoid different variables having same name.
I believe it may trigger some hard-to-debug issues.

scripts/docker-build
hellais: Because `github.com/ooni/explorer` in the future will be `oo/explorer`, but currently is pushed to `oo/explorer-next` to avoid conflicting with the legacy explorer.

ansible/roles/ooni-explorer-next/defaults/main.yml
hellais: What is BCP? You mean that this is the same variable name as the `ooni-explorer` playbook?

I guess we can change it, but eventually one of these two playbooks will be dropped and only the next gen explorer will be kept.

scripts/docker-build
darkk: So the story is that both `TheTorProject/ooni-explorer` and `ooni/explorer` become `openobservatory/explorer`, right?

Is it actually an issue? TBH, I'm 96% okay with both repositories building images with `openobservatory/explorer` name. Anyway, the deploy script uses tag and tag has date and commit hash, I don't expect any conflicts there.

If you still consider that an issue, can you rather specify exact name in the script instead of using some environment variable from some `.profile`?

One of core reasons to create this repo was to extract all the needed bits of knowledge from individual developers' laptops into shared repo and this `IMG_SLUG` override conflicts with that goal.

hellais: > If you still consider that an issue, can you rather specify exact name in the script instead of using some environment variable from some .profile?

I don't think I understand what you mean by this. Does docker-build support passing a custom name for the IMG_SLUG?

ansible/roles/ooni-explorer-next/defaults/main.yml
darkk: Best current practice. Yes, I mean that there is same variable name in `ooni-explorer`. I'm sure, it'll not trigger any issues right now, I highlight that mostly for grepability reasons. Please, change names, if that's not too complex.

Also, different services should use different network endpoints. Imagine a case when you have to move a service to another VM in a rush and you face IP conflict as that IP is already present on docker network there.
So, the subnet of backend_ipv4 should be updated to some other [random subnet](https://github.com/ooni/sysadmin/blob/master/scripts/network-endpoint), if you think that it's reasonable to follow that practice.

scripts/docker-build
darkk: > Does docker-build support passing a custom name for the IMG_SLUG

No. And I think that it should not as IMG_SLUG value is not stored in this repo.

If you want to override image name for `ooni/explorer` repo, then, please, change the script to specify the name for that repo instead of introducing some external state to the script, e.g. parsing `git rev-parse --show-prefix` before it's stripped.

hellais: You mean adding to this script a line similar to:

```
if [ $repo -eq "ooni/explorer"];then
img_slug="ooni/explorer-next"
fi
```

?

darkk: Yep. So anyone running the script in the root of the repo will get approximately same results (modulo state of unpinned packages fetched from external repos).

hellais: SGTM

Issue #241, opened 2018-11-10
WDC migration and re-numbering


darkk:
GH / WDC migrates to Miami.
Hardware will be physically moved, IP addresses will be re-numbered.
So, everything that is there, should be killed, frozen or moved.
This migration will happen in the week 26 Nov ... 02 Dec.

- [x] munin.ooni.io
- [x] shwdc.ooni.io
- [x] shark.ooni.nu
- [x] a.echo.th.ooni.io
- [x] c.web-connectivity.th.ooni.io
- [x] b.collector.ooni.io
- [x] analytics.ooni.io

darkk:
- munin &mdash; gone, disk image will be removed in a while, Tor TH will be re-deployed preserving keys
- shwdc &mdash; gone, replaced with another bastion host, disk image is deleted as it had no files in home directories

darkk:
- shark.ooni.nu &mdash; VM deleted, but disk image is kept for a while. I believe, this VM was used for one-off testing and it has not seen any shell logins since Oct 16, so I'm unsure if allocated resources are still needed. @hellais, please tell, if you still need the data stored there.

darkk:
## analytics.ooni.io

I believe, it can go down for a week and come back online after migration. Moreover, it uses only 1.2 GB of RAM (470 MB without disk cache) out of 6 GB allocated.

darkk:
# c.web-connectivity.th.ooni.io & b.collector.ooni.io

I decided to avoid migration of these services, so I just disabled corresponding collector and test-helper at the bouncer.

darkk:
`b.collector.ooni.io` is hardcoded, so bouncer reply is probably not respected (anyway, traffic did not go away after bouncer configuration change) -- https://github.com/measurement-kit/measurement-kit/blob/v0.9.0-beta/src/libmeasurement_kit/ooni/collector_client.hpp#L18-L19
Seems, endpoint should be migrated somewhere.

darkk:
UPD from @bassosimone: the collector is indeed hardcoded, but the hardcoded value is used as a fallback when bouncer does not reply. It explains both the existence of traffic and very low amount of measurements hitting `b.collector.ooni.io` endpoint.

darkk:
Unfortunately `c.web-connectivity.th.ooni.io` still gets some traffic to `y3zq5fwelrzkkv3s.onion`. The source is unclear.

darkk:
[...]

PR #247, opened 2018-11-25
WDC sunset

darkk:
Some code related to the issue #241 (WDC migration and re-numbering).


darkk:
Inventory is in known-broken state as I plan DNS changes when VMs finally go down.

ansible/roles/ooni-bouncer/tasks/main.yml
```diff
@@ -24,9 +24,10 @@
 
 - name: Configure the bouncer file
   template:
-    src: bouncer.yaml.j2
+    src: bouncer.metayaml
```
hellais: I believe the ansible convention for templates is to make them end in `.j2`. Is there a reason to not follow it in this case?

ansible/roles/ooni-bouncer/templates/bouncer.metayaml
```diff
@@ -0,0 +1,2 @@
+{# meta-template to ensure that the templated configuration file is at least valid yaml #}
+{{ lookup('template', 'bouncer.yaml.j2') | from_yaml | to_nice_yaml }}
```
hellais: I now understand why above you call it `metayaml` instead of `.j2`

ansible/roles/ooni-bouncer/templates/bouncer.yaml.j2
```diff
@@ -15,9 +21,13 @@ collector:
...
-    test-helper: {dns: '37.218.247.110:57004', http-return-json-headers: 'http://37.218.247.95:80',
-      ssl: 'https://37.218.247.110', tcp-echo: 37.218.247.110, traceroute: 37.218.247.110,
-      web-connectivity: 'httpo://y3zq5fwelrzkkv3s.onion'}
+    test-helper:
+      dns: '{{ c_echo_th }}:57004'
+      http-return-json-headers: 'http://{{ a_http_th }}:80'
+      ssl: 'https://{{ c_echo_th }}'
+      tcp-echo: {{ c_echo_th }}
+      traceroute: {{ c_echo_th }}
+      web-connectivity: 'httpo://y3zq5fwelrzkkv3s.onion'
```
hellais: 👍

ansible/roles/ooni-bouncer/tasks/main.yml
darkk: `.j2` makes things actually slightly more painful for me: my editor disables meaningful syntax highlighting for `*.j2` files. E.g. blah.yaml.j2 does not look like yaml file for me. That's why I never used `.j2` suffix. :)

ansible/roles/ooni-bouncer/templates/bouncer.metayaml
darkk: I just did not care about extension for the file that has just two lines, so I just added something that is one bit better than `bouncer.blah`. :man_shrugging: 

The only reason for me to specify extensions in those templates is to get syntax highlighting for the "base-format".

PR #249, opened 2018-11-27
Install python-yaml to unbrick the daily cronjob

hellais:
This actually does not work on the latest GH machine, because the base system is missing `python-apt`


hellais:
see: https://github.com/ooni/sysadmin/issues/250

PR #251, opened 2018-11-28
Fix systemd alerts

hellais:
I believe now the only one that is still broken is `bouncer.infra.ooni.nu`, yet this is a box we should phase out/deprecate anyways and `labs.ooni.io` where the certbot script is not working as expected (it's a testing machine so it's not essential).

* apparmor was actually only a problem on wiki.ooni.io, which I just disabled
* There was a certbot problem on hkgbouncer, which was due to some empty config file being left over from some bad upgrade or something (I just deleted it)
* datacollector had a broken ntp service, which I fixed by running updating the ntp client

The remaining alerts were just related to smartd being broken, which is now disabled as part of the playbook on all greenhost boxes.

### DIFF
--- a/ansible/dom0-bootstrap.yml
+++ b/ansible/dom0-bootstrap.yml
@@ -54,3 +54,9 @@
       tags: on-rename
     - role: node_exporter
       tags: node_exporter
+
+- hosts: gh
+  gather_facts: false
+  roles:
+    - role: eclipsis
+      tags: gh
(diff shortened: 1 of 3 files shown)


ansible/dom0-bootstrap.yml
```diff
@@ -54,3 +54,9 @@
       tags: on-rename
     - role: node_exporter
       tags: node_exporter
+
+- hosts: gh
+  gather_facts: false
+  roles:
+    - role: gh
+      tags: gh
```
darkk: I'd suggest to move those bits right after `dom0-defaults`. That order feels a bit more logical.
If there is a reason for this order, I'd suggest to add a comment.

hellais: There is no reason to have it her. I will move it up.

PR #253, opened 2018-12-06
Bump orchestra-scripts to a working version

hellais:
See: https://github.com/ooni/orchestra/pull/57 for details on the bug

### DIFF
--- a/ansible/roles/airflow/files/docker-trampoline
+++ b/ansible/roles/airflow/files/docker-trampoline
@@ -196,7 +196,7 @@ case "$script" in
     volargs="${volargs} --tmpfs /tmp:rw,noexec,nosuid,size=64m"
     set -o xtrace
     exec docker run $docksafe --user "$uidno" $docknet ${volargs} \
-      openobservatory/orchestra-scripts:20181022-1f800a5b \
+      openobservatory/orchestra-scripts:20181206-3be1378e \
       sync-test-lists.py \
         --working-dir /tmp \
         --postgres 'host=db-1.proteus.ooni.io user=proteus dbname=proteus sslmode=verify-full sslrootcert=/ooca-cert/postgres_ca.cert'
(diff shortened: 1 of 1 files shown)


darkk:
We have airflow 1.8.0. The DAG in question is [configured](https://github.com/ooni/sysadmin/blob/70672fec6b744df716e106a4ceccb9ecd5e62753/ansible/roles/airflow/files/airflow-dags/sync.py#L8-L10) with `start_date=2018-10-21`, `schedule_interval=1h` and `catchup=False`.
The behavior we observe is following: when the DAG was paused for a while and it was unpaused after that, then the scheduler does not wait for precise `schedule_interval` to trigger DAG, but triggers that immediately. It makes sense as soon as _"catchup"_ is disabled, so per-bucket backfilling is not required.
I considered that airflow bug because:
- some of task instances were unable to display their status page and logs: airflow webserver was presenting a mushroom cloud (internal server error page)
- DAG _"Tree view"_ was unable to show the history for the DAG, only the DAGs after the "incident" were shown, but the history was gone

I've deleted some* TIs**, it did not help &mdash; the airflow webserver was still crashing, logs still were not shown, _"Tree view"_ was broken.

*) one or two, don't remember exactly, that can be probably reconstructed with logs
**) task instances, that's airflow term

So I've dug into the airflow code & relevant discussions &mdash; everything seemed fine. So I've put the task aside for a while, going to follow "drop the DAG state and start it from scratch" short-cut route. Now, before taking the route (and backing up the airflow database) I've looked at views and state of TIs for the DAG. And everything is fine: logs are served, TIs display their status, _"Tree View"_ shows reasonable history.

The puzzle is not solved, but the airflow self-healed in some way and I don't think that investing more time makes sense. Maybe, the crash I saw after TI cleanup was caused by some erratic data that was cached somewhere...

So I'm just patching the script to accept the fact that start/end date may have microseconds, and that's it.

PR #255, opened 2018-12-10
Purge ntp from hosts

hellais:


ansible/roles/ntp/tasks/main.yml
```diff
@@ -18,4 +18,10 @@
 - name: configure chrony
   template: src=chrony.conf dest=/etc/chrony/chrony.conf owner=root group=root mode=0444
   notify: restart chrony
+
+- name: purge ntp to avoid ntp.service from running
+  apt:
+    name: ntp
+    state: absent
+    purge: yes
```
darkk: I wonder, wasn't it enough to add `purge: yes` to `drop common ntpd packages` step?

hellais: Probably, but that also includes the purge of some other unrelated package.

PR #256, opened 2018-12-10
Serve static files as placeholder API data for explorer country pages

sarathms:
Modified `explorer-next`'s nginx configuration to mock API data from static files stored in `/var/www/explorer/data/`.

### DIFF
--- a/ansible/roles/ooni-explorer-next/defaults/main.yml
+++ b/ansible/roles/ooni-explorer-next/defaults/main.yml
@@ -1,3 +1,4 @@
 explorer_next_tag: '20181112-00674a72'
 explorer_next_backend_ipv4: 172.27.33.254
 explorer_next_backend_port: 3100
+explorer_country_data_path: /var/www/explorer
--- a/ansible/roles/ooni-explorer-next/defaults/main.yml
+++ b/ansible/roles/ooni-explorer-next/defaults/main.yml
@@ -1,4 +1,4 @@
-explorer_next_tag: '20181112-00674a72'
+explorer_next_tag: '20181212-b55c9c80'
 explorer_next_backend_ipv4: 172.27.33.254
 explorer_next_backend_port: 3100
 explorer_country_data_path: /var/www/explorer
--- a/ansible/roles/ooni-explorer-next/templates/ngx-explorer
+++ b/ansible/roles/ooni-explorer-next/templates/ngx-explorer
@@ -26,8 +26,9 @@ server {
     access_log  /var/log/nginx/{{ explorer_domain }}.access.log; # FIXME: log_format
     error_log   /var/log/nginx/{{ explorer_domain }}.error.log;
 
+    # XXX: Temporary development setup to be replaced by api.ooni.io
     location /data/ {
-        root {{ explorer_country_data_path }};
+        root {{ explorer_country_data_path }}/;
     }
 
     location / {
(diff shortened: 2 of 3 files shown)


ansible/roles/ooni-explorer-next/templates/ngx-explorer
```diff
@@ -26,6 +26,10 @@ server {
     access_log  /var/log/nginx/{{ explorer_domain }}.access.log; # FIXME: log_format
     error_log   /var/log/nginx/{{ explorer_domain }}.error.log;
 
+    location /data/ {
+        root {{ explorer_country_data_path }};
+    }
+
```
darkk: FYI.
I understand the convenience of that during development (and that's, unfortunately, the only way to do that right now). But also, I would advice against doing alike stuff on the production explorer setup.
I'd suggest permanent caching of specific URL prefix instead (with versioned URLs), or non-permanent caching and cache destruction on roll-out.
The reason is: it's way easier to build, roll-out and roll-back the container as a whole, than to have some bits of stateless web-app in the container and some bits on the filesystem.
Please, document that it's dev-only crutch in the config file If you agree with that, so it does not end up in the production setup "by accident". If you disagree, let's discuss :)

darkk: Also, beware of [off-by-slash](https://twitter.com/x0rz/status/1052899891624710145). :)

hellais: Yes I agree this is a dev only "crutch". I think adding a simple comment in the ngx-explorer config file saying "# XXX this is for development only".

ansible/roles/ooni-explorer-next/defaults/main.yml
```diff
@@ -1,3 +1,4 @@
 explorer_next_tag: '20181112-00674a72'
 explorer_next_backend_ipv4: 172.27.33.254
 explorer_next_backend_port: 3100
+explorer_country_data_path: /var/www/explorer
```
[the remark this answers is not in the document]
sarathms: Done. Do I also have to setup copying all the files in the directory from somewhere? Or is this just to prevent nginx from trying to serve from a non-existent directory somehow?

ansible/roles/ooni-explorer-next/templates/ngx-explorer
sarathms: Wow. I had to decide whether the slash is necessary or not and ended up dropping it after looking at all the examples from NGINX documentation. None of them use a `/` for directory paths.

Added the comment too.

ansible/roles/ooni-explorer-next/defaults/main.yml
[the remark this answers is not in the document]
hellais: I think it's fine that you do that manually as that is part of the development process. The reason to create the dir as part of the playbook is to ensure that on a clean machine it will do everything that is necessary to get it to work (manually settings permissions or creating directories can be error prone).

ansible/roles/ooni-explorer-next/templates/ngx-explorer
darkk: @sarathms I'm unsure if it applies to `root`, or just to `alias`. I don't have in-deph understanding of that vulnerability.

Issue #175, opened 2017-10-11
Cleanup GH hosts that are not in inventory


darkk:
Following hosts are in `gh-droplets.json` and are NOT in `inventory` right now.

- [x] `{"v4":"199.119.112.12","name":"test-qemu.infra.ooni.io","dc":"Washington DC"}`
- [x] `{"v4":"37.218.240.117","name":"design.ooni.io","dc":"Hong Kong"}`
- [x] `{"v4":"37.218.240.156","name":"ooniprobe-deb-testing","dc":"Hong Kong"}`
- [x] `{"v4":"37.218.240.158","name":"afwrk01","dc":"Hong Kong"}`
- [x] `{"v4":"37.218.240.159","name":"afwrk02","dc":"Hong Kong"}`
- [x] `{"v4":"37.218.240.160","name":"afwrk03","dc":"Hong Kong"}`
- [x] `{"v4":"37.218.240.162","name":"afwrk04","dc":"Hong Kong"}`
- [ ] `{"v4":"37.218.240.93","name":"wiki.ooni.io","dc":"Hong Kong"}`
- [x] `{"v4":"37.218.242.175","name":"amsmetadb.infra.ooni.io","dc":"Amsterdam"}`
- [x] `{"v4":"37.218.242.178","name":"dataproxy.infra.ooni.io","dc":"Amsterdam"}`
- [x] `{"v4":"37.218.242.28","name":"test-ooniprobe-trusty","dc":"Amsterdam"}`
- [x] `{"v4":"37.218.242.32","name":"ooniprobe-test-xenial","dc":"Amsterdam"}`

afwrk{01..04} is my experiment with distributed parallel backfill (to use more CPUs).
amsmedadb, dataproxy are resources allocated for parts of new pipeline.
What's the destiny of other VMs?

hellais:
> wiki.ooni.io

Is a wiki I had setup for sketching out the contextual database project. Maybe it's not the optimal wiki solution, but @agrabeli should probably comment on that.

> design.ooni.io

Is actually not booting due to eclipsis not having resources when I started it. I guess it should be killed and restarted. I plan to move the oo/design publication from github to there.

> test-qemu.infra.ooni.io
> ooniprobe-deb-testing
> test-ooniprobe-trusty
> ooniprobe-test-xenial

no idea what these are, but since some are in AMS we should consider cannibalising them for some other useful purposes.

bassosimone:
[...]

PR #275, opened 2019-02-20
Tor TH

darkk:
That:
1. migrates Tor TH from WDC to MIA
2. drops wiki.ooni.io VM that was used at HKG Tor TH (disk named `wiki.ooni.io:disk0` is preserved for a while)
3. deploys hkgtorth.ooni.nu instead of wiki.ooni.io
4. re-deploys amstorth with clean disk image (old one named `ooni-measurements-productions:disk0` from ancient measurements.ooni.io is preserved for a while)


darkk:
Noticed #175, moving on.

Issue #276, opened 2019-02-20
 GH hosts missing from inventory


darkk:
@hellais @bassosimone what's the destiny of the following hosts? Should they be kept? Should they be dropped?

- [ ] `{"v4":"103.104.244.21","name":"testing","dc":"Hong Kong"}`
- [ ] `{"v4":"37.218.241.4","name":"cu-testing-s1","dc":"US East (Miami)"}`
- [ ] `{"v4":"37.218.242.89","name":"cu-testing-s2","dc":"Amsterdam"}`

They were created some time ago, but they're not tracked in `inventory`, not monitored and so on.

bassosimone:
@darkk I do not know anything about them so I’d say deploy the napalm, then maybe we can go windsurfing :)

hellais:
cu-testing-* where some boxes I setup to do some experiments, but didn't get too far with the experimentation.

testing (according to my ~/.ssh/config) is marked as `dns.ooni.io`, but I don't recall what it was for.

I suggest dropping all three of these hosts and if we need them they can be recreated.

darkk:
Okay. I stopped them and will drop tomorrow. `testing` was unable to boot properly anyway, it's console was showing busybox.

PR #277, opened 2019-02-21
Update ooni-collector docker image

hellais:
For deploying: https://github.com/TheTorProject/ooni-backend/pull/117


hellais:
I am now trying to roll this out in collectora, which is not being used by any probes out there as test.

hellais:
After having rolled this out to a.collector.ooni.io I have tested the following:

* measurement_kit 0.9.0 run with: `./measurement_kit --collector https://a.collector.ooni.io http_invalid_request_line` is able to successfully submit a measurement to the collector
* using go-measurement-kit and making tweaks to `web_connectivity.go` if the software_name is `ooniprobe-android` and the version is `2.0.0` the measurement is rejected (mk emits a failure `failure.measurement_submission`)
* using go-measurement-kit, with similar tweaks to `web_connectivity`, and `ooniprobe-android 2.0.1`, the measurement upload works as expected

Below is the extract from `web_connectivity.go`:
```
	nt.Options = mk.NettestOptions{
		IncludeIP:        false,
		IncludeASN:       true,
		IncludeCountry:   true,
		DisableCollector: false,
		CaBundlePath: "testdata/cert.pem",
		CollectorBaseURL: "https://a.collector.ooni.io",
		SoftwareName: "ooniprobe-android",
		SoftwareVersion: "2.0.0",
		GeoIPCountryPath: "",
		GeoIPASNPath:     "",
		OutputPath:       "/tmp/ooniprobe-report.jsonl",
		LogLevel:         "INFO",
		Inputs:           []string{"https://ooni.torproject.org/"},
	}
```

Based on the above testing I believe it's reasonable to deploy it to b.collector & c.collector prior to the announcement post.

I would be useful to have somebody else sanity check this before I roll it out.

@darkk @bassosimone

ansible/inventory-check.yml
```diff
@@ -129,7 +129,7 @@
 
     - name: stamp the inventory that was checked
       assert:
-        that: "'{{ inventory_hash }}' == '0aeedabeac0fcdf921b952ead8f920b847e7041aae62d7cc52d52184210aea3c'"
+        that: "'{{ inventory_hash }}' == '390072320b73bf063d6602bfb4545329099cb8e6d1487a26b6d4c28f7dbc7aae'"
```
darkk: How is it possible that the check passed? It failed due to #276 for me.

hellais: Not sure. However the inventory check was failing in master as well.

darkk:
> deploy it to b.collector & c.collector prior to the announcement post.

I'd suggest to roll out to any of these two, take a look (e.g. check one bucket or over the weekend) and then try another one.

I believe, [deploy-collector.yml](https://github.com/ooni/sysadmin/blob/master/ansible/deploy-collector.yml#L12-L18) does not list `b` and `c` for a reason, the recipe may be unsuitable for them as-is, maybe something has to be tuned / changed.

Also, onion keys should be kept in mind.

hellais:
> I'd suggest to roll out to any of these two, take a look (e.g. check one bucket or over the weekend) and then try another one.

Yeah that is a good idea.

> I believe, [deploy-collector.yml](https://github.com/ooni/sysadmin/blob/master/ansible/deploy-collector.yml#L12-L18) does not list `b` and `c` for a reason, the recipe may be unsuitable for them as-is, maybe something has to be tuned / changed.

You are right in this point. I believe the reason for that was that b & c did not require a redeploy and to avoid downtime we decided not to risk it.

This is a good opportunity to check that the deploy scripts actually work properly on those hosts too.

hellais:
I have deployed this to b.collector.ooni.io (aka miacollector.ooni.nu).

In order to get it to work I had to make some changes to the playbook and rerun `dom0-bootstrap.yml` on the host.

I also had to manually terminate the old docker image running on the host.

I did a manual smoke test, similar to the above, to validate that everything was working as expected (initially it was still accepting measurements from 2.0.0 and that was due to the conflicting old docker container).

If everything has worked OK on monday I say we do the same for c.collector.

I do see some failures in #ooni-bots related to `Last `collector_sensor_miacollector` failure 1d 0h ago.
Last `rsync_collector_miacollector` failure 1d 0h ago.
and will look into these too.

darkk:
> In order to get it to work I had to make some changes to the playbook and rerun `dom0-bootstrap.yml` on the host.

Unfortunately, it does not explain what was going wrong.

> I do see some failures in #ooni-bots ... and will look into these too.

As soon as 8h passed by and it's not resolved, I assume, I can take over without stepping on your toes.

The [airflow log](http://127.0.0.1:8080/admin/airflow/log?task_id=collector_sensor_miacollector&dag_id=fetcher&execution_date=2019-02-21T00:00:00) clearly states SSH authentication failure:

```
[2019-02-22 00:00:15,340] {base_task_runner.py:95} INFO - Subtask: + rsync miacollector.ooni.nu:
[2019-02-22 00:00:15,341] {base_task_runner.py:95} INFO - Subtask: ++ wc -l
[2019-02-22 00:00:15,722] {base_task_runner.py:95} INFO - Subtask: Could not create directory '/home/ooshovel/.ssh'.
[2019-02-22 00:00:17,687] {base_task_runner.py:95} INFO - Subtask: Permission denied (publickey).
[2019-02-22 00:00:17,688] {base_task_runner.py:95} INFO - Subtask: rsync: connection unexpectedly closed (0 bytes received so far) [Receiver]
[2019-02-22 00:00:17,688] {base_task_runner.py:95} INFO - Subtask: rsync error: unexplained error (code 255) at io.c(226) [Receiver=3.1.1]
```

The issue is that UID/GID was rewritten by dom0-bootstrap playbook as it does not take into account that the user may already exist with "historical" UIDs (all in all, "bootstrap" implies clean initial state):

```
Feb 22 00:35:15 b groupmod[22447]: group changed in /etc/group (group sync-user/1000, new gid: 8002)
Feb 22 00:35:15 b groupmod[22447]: group changed in /etc/passwd (group sync-user/1000, new gid: 8002)
Feb 22 00:36:24 b usermod[22833]: change user 'sync-user' UID from '1000' to '8002'
Feb 22 00:36:24 b usermod[22833]: change user 'sync-user' home from '/home/sync-user' to '/nonexistent'
Feb 22 00:36:24 b usermod[22833]: change user 'sync-user' shell from '/bin/bash' to '/usr/sbin/nologin'
Feb 22 00:50:43 b usermod[24286]: change user 'sync-user' home from '/nonexistent' to '/home/sync-user'
Feb 22 00:50:43 b usermod[24286]: change user 'sync-user' shell from '/usr/sbin/nologin' to '/bin/bash'
```

Shell was changed back, but UID was not. So the `sync-user` probably was unable to use `authorized_keys` or some other file that is owned by 1000:1000. So `sudo chown -R sync-user:sync-user /home/sync-user` fixed the issue.

hellais:
I am now deploying this to c.collector

hellais:
I have deployed this on c.collector.ooni.io

I did the following:
• Ran `dom0-bootstrap` script
• Added entry and redeployed
• Fixed permissions on sync-user home as suggested above by @darkk 
• Did a quick inspection that all the permissions looked in order
• Ran the smoke test to vaildate ooniprobe-android measurements from 2.0.0 are not collected and normal ones are

I believe it should be properly deployed, but perhaps @darkk you can do another quick check tomorrow to ensure everything is in order and the spice is flowing?

darkk:
> If all looks good let's merge this PR and consider it done.

As we've discussed quite some time ago, the infrastructure has no "branches", PR's are nothing but forums for discussions and any PR should be self-merged soonish after deployment :-)

> quick check

Will check stats in a couple of pipeline cycles (on Monday).

darkk:
@hellais I believe, something went wrong somewhere.

```
5.5G	2019-02-01
5.7G	2019-02-02
7.1G	2019-02-03
5.8G	2019-02-04
5.9G	2019-02-05
6.3G	2019-02-06
6.2G	2019-02-07
5.7G	2019-02-08
5.6G	2019-02-09
5.3G	2019-02-10
5.2G	2019-02-11
5.7G	2019-02-12
5.3G	2019-02-13
5.7G	2019-02-14
5.3G	2019-02-15
5.2G	2019-02-16
5.9G	2019-02-17
4.8G	2019-02-18
4.8G	2019-02-19
5.3G	2019-02-20
7.8G	2019-02-21
4.9G	2019-02-22
4.8G	2019-02-23
4.5G	2019-02-24
4.7G	2019-02-25
4.9G	2019-02-26
4.8G	2019-02-27
4.7G	2019-02-28
1.3G	2019-03-01
```

darkk:
I'd say, b.collector also [changed behavior](https://mon.ooni.nu/grafana/dashboard/db/node-exporter-full-single-machine?orgId=1&from=now-14d&to=now&var-node=miacollector.ooni.nu&var-port=9100&refresh=1m) and size of "daily batch" noticeably after deployment on the 22th.

b collector 2019-03-02 22-48-58

hellais:
I took a quick look at the distribution of reports submitted by ooniprobe version and it seems like it's affecting the python ooniprobe versions. It seemed like older version (like 1.3.2 & 1.5.0) were especially affected (these versions don't do collector failover).

I tried reaching the onion service of the c.collector and it appeared to not be working. I believe the . problem was bad permissions on the onion service private_key and hostname file.

I did: `chown -R collector:collector /data/*/tor/` and restarting the docker service on both b.collector & c.collector.

Let's see if this makes the stats resume to normal operation.

hellais:
Looking at the charts and stats today, it seems to me like it has resumed normal operation.

screenshot 2019-03-04 at 08 50 56
screenshot 2019-03-04 at 08 50 29


It's not so nice that ooni-backend does not trigger any error when it fails to start the tor hidden service.
Since it appears like we rely quite heavily on these HS working as expected, we should perhaps find a way to monitor their accessibility via prometheus.

hellais:
It seems like the bucket sizes are now back to a reasonable size:
```
1.3G	2019-03-01
2.6G	2019-03-02
4.3G	2019-03-03
7.3G	2019-03-04
5.1G	2019-03-05
5.3G	2019-03-06
5.4G	2019-03-07
```

I am closing and merging this.

PR #280, opened 2019-02-22
Increase the max URI size for OONI Run

hellais:


darkk:
Sidenote: I'd suggest to consider an idea to create a custom [`error_page`](http://nginx.org/en/docs/http/ngx_http_core_module.html#error_page) for 414.

PR #284, opened 2019-03-13
Add miaapi.ooni.nu

hellais:


hellais:
Greenhost seems to have changed the base image settings so I have had to manually change the iptables rules with: `iptables -P INPUT ACCEPT`

We should probably codify this inside of the dom0 script so that it happens for every host.

PR #285, opened 2019-03-19
Use test API server for Explorer test instance

sarathms:
I tried switching the API server address provided as an environment variable. Doesn't seem to be the right way. The country page which fetches data from the new endpoints on the API server seems to still hit the production API server (based on docker logs).


sarathms:
The test APIs have been deployed to production. Explorer now uses the production endpoints directly.

PR #288, opened 2019-04-04
Make sure @bassosimone keys are current

bassosimone:


bassosimone:
> Don't forget to roll-out new keys.

Do you mean that I should `ansible play` all?

darkk:
Yep.
I mean, I can do it as well, but, probably, ensuring that everything works for you for all the hosts in `inventory` is a useful exercise.

bassosimone:
I agree. Will do on Monday (if something goes wrong, I may not be able to fix stuff easily and I’d rather not waste the weekend.)

PR #290, opened 2019-04-08
Add sshd AllowUsers

darkk:
To make ACL a bit more strict.

I'm also going to check if `root` was used for logins recently (within `/var/log/auth.log*` rotation time window) and drop it if it was not.


darkk:
That was rolled out with...
```
$ ./play dom0-bootstrap.yml -t adm,sshd --start-at-task='adm : create .ssh/authorized_keys' 
```

PR #291, opened 2019-04-08
Enforce /root/.ssh/authorized_keys in addition to authorized_keys2

darkk:
`/root/.ssh/authorized_keys` included various keys of OONI Weasel, @willscott, @joelanders and some others, so it's cleaned up.

The keys that did not look like some key of "gone" team member (and could be relevant to some process) were added to  `root_ssh_authrorized_keys` variable.

@bassosimone @hellais you may want to review those keys and/or revoke some of them.


bassosimone:
Thanks @darkk. I don't know any of the keys listed there.

PR #292, opened 2019-04-10
Add support for generating user certificates

hellais:
Related to https://github.com/ooni/sysadmin/issues/272


ansible/roles/plpsql/templates/pg_hba.conf
```diff
@@ -92,9 +92,11 @@ host    all             all             ::1/128                 trust
 #host    replication     postgres        127.0.0.1/32            trust
 #host    replication     postgres        ::1/128                 trust
 
-# NB: `ssl` is neither supported nor enforced.
 {% if inventory_hostname == 'hkgmetadb.infra.ooni.io' %}
-host replication amsrepl {{ lookup('dig', 'amsmetadb.ooni.nu/A') }}/32 md5
+hostssl replication amsrepl {{ lookup('dig', 'amsmetadb.ooni.nu/A') }}/32 cert
```
darkk: What CA is used to verify the client? As far as I see, `ssl_ca_file` is not set.

Also, CRL (`ssl_crt_file`) is not set, but revocation was not considered before as, given the low number of clients and lack of HA requirements, it may be easier to roll-over CA key in case of the key being compromised.

ansible/roles/plpsql/templates/recovery.conf
```diff
@@ -1,4 +1,4 @@
 standby_mode = 'on'
 primary_slot_name = 'amsmetadb_3'
-primary_conninfo = 'host=hkgmetadb.infra.ooni.io port=5432 user=amsrepl password={{ amsrepl_hkgmetadb_password }}'
+primary_conninfo = 'host=hkgmetadb.infra.ooni.io port=5432 user=amsrepl sslcert=/srv/pl-psql/postgres_amsrepl_user.cert sslcert=/srv/pl-psql/postgres_amsrepl_user.key'
```
darkk: Is `.vault` suffix auto-stripped?

ansible/deploy-pipeline.yml
```diff
@@ -23,7 +23,7 @@
       login_password: '{{ shovel_postgres_password }}'
       db: metadb # otherwise connection fail
       name: amsrepl
-      password: '{{ amsrepl_hkgmetadb_password }}'
+      password: 'INSECURE_USE_SSL_CERT'
```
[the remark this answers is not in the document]
hellais: Probably not, though I recall we were discussing of setting it explicitly to some value to make it clear that it should not be relied upon.

We can perhaps remove it though.

ansible/files/Makefile
```diff
@@ -35,5 +35,13 @@ darkk_weechat.cert: darkk_weechat.key pusher_ca.cert pusher_ca.key.vault
...
+
 %.key:
 	touch $@ && chmod 600 $@ && openssl genrsa 2048 >$@
+
+%_user.key.vault:
+	openssl genrsa 2048 | ../vault encrypt --output $@
+
+postgres_amsrepl_user.cert: postgres_amsrepl_user.key.vault postgres_ca.cert
+	openssl req -new -key <(../vault view postgres_amsrepl_user.key.vault) -sha256 -days 3650 -subj '/O=OONI/OU=AMS Replica User/CN=amsrepl' | \
+	openssl x509 -req -sha256 -days 3650 -CA postgres_ca.cert -CAkey <(../vault view postgres_ca.key.vault) -set_serial 0x`head -c 16 /dev/urandom | xxd -plain` -out $@
```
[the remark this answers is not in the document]
hellais: I figured that since that CA is already going to be distributed to all the pg instances we may as well use the same one for also signing client certificates.

Do you think that is a bad idea?

ansible/files/postgres_amsrepl_user.key.vault
```diff
@@ -0,0 +1,89 @@
...
+33353166646464383234663063613137333135356163613834366630346363326534393934323163
+30356163363735313631633163653232613764336436343961653165333132363436663231363631
+37393665623464346363366233663136396237393430386362646133333363663265396264396163
+30313034363639346337306562326562343338303161613762383930343339616561313966303966
+37313762656264643035363639663562333431356536616337343938363766623836366430646634
+38626537373333643033363064383037653539356335656561323832326461666138346138646561
+37343535613166383637353861343235656535353833343033303733666433396235623865343535
+37663233366239366330323333383838323837343231656234383130326264316131316536306230
+34313137646133643765313365653638623832313133323936623765623465316532373463646631
+39356465383734633432
```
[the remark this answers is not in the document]
hellais: Hum. Can you show me a place where another pattern for this is done?

I followed what was being done for other key files in `ansible/files/*.vault`.

ansible/roles/plpsql/templates/pg_hba.conf
hellais: Ah good point. I will add that.

I agree that revocation is currently out of scope.

ansible/roles/plpsql/templates/recovery.conf
hellais: My understanding is that the copy command will transparently decrypt a vault file and copy it unencrypted (see: https://github.com/ansible/ansible/pull/15417).

ansible/deploy-pipeline.yml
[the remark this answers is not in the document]
darkk: Yep, I suggested to have IP-limited usernames with insecure passwords and you suggested to drop the passwords altogether with `PASSWORD NULL` and replace them with certificates, to avoid any sort of secrets being stored in the PG itself.
That's why I was surprised to see insecure password instead of NULL password :-)

ansible/roles/plpsql/templates/recovery.conf
darkk: It will decrypt, yes. Will it strip `.vault` suffix from the filename? :)

hellais: Hum I see what you mean. Yeah I suspect you are right that it will have the vault prefix because the `dest` argument has the item name with the `vault` suffix. I will look into fixing that.

ansible/roles/plpsql/templates/recovery.conf
```diff
@@ -1,4 +1,4 @@
 standby_mode = 'on'
 primary_slot_name = 'amsmetadb_3'
-primary_conninfo = 'host=hkgmetadb.infra.ooni.io port=5432 user=amsrepl password={{ amsrepl_hkgmetadb_password }}'
+primary_conninfo = 'host=hkgmetadb.infra.ooni.io port=5432 user=amsrepl sslcert=/srv/pl-psql/pg_user_amsrepl.cert sslcert=/srv/pl-psql/ssl/pg_user_amsrepl.key'
 trigger_file = '/srv/pl-psql/trigger.flag'
```
darkk: You may want to add `sslmode=verify-full` here.
[_By default, PostgreSQL will not perform any verification of the server certificate._](https://www.postgresql.org/docs/9.6/libpq-ssl.html)

ansible/roles/plpsql/templates/postgresql.conf
```diff
@@ -76,14 +76,14 @@ max_connections = 100			# (change requires restart)
...
-#ssl_cert_file = 'server.crt'		# (change requires restart)
-#ssl_key_file = 'server.key'		# (change requires restart)
-#ssl_ca_file = ''			# (change requires restart)
+
+ssl = on
+ssl_ciphers = 'ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-SHA384:ECDHE-RSA-AES256-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256'
+ssl_prefer_server_ciphers = on
+ssl_cert_file = '/srv/pl-psql/ssl/{{ inventory_hostname }}.cert'
+ssl_key_file = '/srv/pl-psql/ssl/{{ inventory_hostname }}.key'
+ssl_ca_file = '/srv/pl-psql/ssl/pg_user_ca.cert'
```
darkk: I'd say, it looks like a bad idea to put secret key in PGDATA as soon as we're going to upload snapshot of PGDATA directory to S3 monthly.

hellais: Aren't we going to be uploading only certain files from that directory?

In any case this approach of putting a subdirectory in there turned out to not work for other reasons as well so it's changed in the latest commit.

ansible/roles/plpsql/templates/recovery.conf
hellais: Good point.

ansible/roles/plpsql/templates/recovery.conf
```diff
@@ -1,4 +1,4 @@
 standby_mode = 'on'
 primary_slot_name = 'amsmetadb_3'
-primary_conninfo = 'host=hkgmetadb.infra.ooni.io port=5432 user=amsrepl password={{ amsrepl_hkgmetadb_password }}'
+primary_conninfo = 'host=hkgmetadb.infra.ooni.io port=5432 user=amsrepl sslcert=/srv/psql-ssl/pg_user_amsrepl.cert sslcert=/srv/psql-ssl/pg_user_amsrepl.key'
```
darkk: > `sslcert=/srv/psql-ssl/pg_user_amsrepl.cert sslcert=/srv/psql-ssl/pg_user_amsrepl.key`

Also, probably, specifying `sslcert` twice is not the thing you want :)

hellais:
We ended up going a different route.
Closing this stale PR.

PR #295, opened 2019-04-15
README.md: document that we need dnspython

bassosimone:
I'm unsure whether we need to apply strict checks like we do for
ansible or we're just fine with suggesting pinning.


bassosimone:
Thanks!

PR #296, opened 2019-04-16
bouncer: deploy openobservatory/backend@20190415-65197d29

bassosimone:


bassosimone:
Deployed! After the deployment:

1. the new API work even though there is some duplication in the returned results (I don't know if we need to care about this, but perhaps we do)

2. MK works

3. the new Golang code for MK works with new APIs

4. ooniprobe 2.3.0 works using Tor

5. more checks

Issue #239, opened 2018-10-31
Monitoring epic, v3


darkk:
- [ ] monitor `Oops` and `BUG:` in kernel log #210
- [ ] add regexp for go app crashes in systemd syslog #220
- [ ] monitor nginx error.log #138
- [ ] monitor public endpoints with non-trivial health check #113 #173 #194
- [ ] monitor public webservice health check from another prometheus #174 
- [ ] monitor all network endpoints with blackbox checks -- 373600887c24eab683d3e918fb45b33cd1d720a6
- [ ] postgres replica latency
- [ ] anonymize IPv4 address to /24 and IPv6 to /48 (min routable prefix: [1](https://nsrc.org/workshops/2016/apricot2016/raw-attachment/wiki/Track2BGP/05-BGP-BCP.pdf), [2](https://www.enog.org/wp-content/uploads/presentations/enog-14/4-IPv6-BGP-Tutorial.pdf) while sou, ) -- 7b12322deb8d217e3ad37d6faea14f8d796a20f8, ea1f6a4826f537b9b725c39800c522a666714cc4
- [ ] use [dockerd monitoring](https://docs.docker.com/config/thirdparty/prometheus/) to track fast restarts with `engine_daemon_container_actions_seconds_count{action="start"}` #220
- [ ] monitor number of systemd service restarts-per-minute (needs node_exporter >0.16.0)

darkk:
BTW, it's relatively easy to monitor tor hidden service within current setup. The plan is:
- re-deploy amstorth.ooni.nu from scratch to clean it up (it was ancient ooni measurements dev/test box)
- add DNSPort and TransPort to tor there
- put blackbox_exporter there under separate uid
- redirect blackbox_exporter traffic to tor (either all, or onion-only)

darkk:
Actually, I'm very unsure if all of the above makes sense right now. I've observed that capacity and rare-errors alerts at `#ooni-bots` are often neither acted upon not silenced, so they turn the channel into stream of noise. I assume that adding more of that would be counterproductive with current alerting setup.

So, given recent incidents, I'm limiting it to monitoring of public `*.onion` endpoints of OONI at this moment as it's both not a trivial and needed check.

hellais:
I agree that it's more important to conclude the other pipeline related tasks and the current state of monitoring doesn't warrant for more noise.

Speaking of which @darkk is it possible to reduce the level of noisiness of the onion service alerts?
I have seen them get pretty chatty recently.

[...]

PR #300, opened 2019-04-18
System-wide *.onion support for Prometheus VM

darkk:
That's infra part of _Monitoring public OONI `*.onion` endpoints_ for #239.

### DIFF
--- a/ansible/deploy-tor.yml
+++ b/ansible/deploy-tor.yml
@@ -3,35 +3,9 @@
 
 - hosts: have_tor
   gather_facts: true # to fetch `jessie` codename
+  roles:
+    - role: tor-bin
   tasks:
-    # I'm aware of https://github.com/nusenu/ansible-relayor but it solves way
-    # more complex problem and needs immediate ansible update and have already
-    # dropped support for debian/jessie, that is still used as of 2018-07-21 :-(
-    - name: fetch and install deb.torproject.org signing key
-      apt_key:
-        url: https://deb.torproject.org/torproject.org/A3C4F0F979CAA22CDBA8F512EE8CBC9E886DDD89.asc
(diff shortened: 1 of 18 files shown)


darkk:
That's actually deployed, so, merging.

SuperQ:
Ohh, neat.

ansible/deploy-tor.yml
```diff
@@ -3,35 +3,9 @@
 
 - hosts: have_tor
   gather_facts: true # to fetch `jessie` codename
+  roles:
+    - role: tor-bin
```
bassosimone: So, this is a subroutine call in the Ansible world, eh?

hellais: It's more like a package of stuff that allow you to configure or install some unit of functionality on a system.
See: https://docs.ansible.com/ansible/latest/user_guide/playbooks_reuse_roles.html

darkk: Yes, I usually treat it as a subroutine call :)
Making a "good" and "reusable" package is a complex exercise and sometimes it leads to (possibly, unnecessary) levels of abstractions.

darkk:
@SuperQ yep. I've tested it within another infra for a while and it works quite well, it also helped to understand some issues regarding hidden service management in [dynamic](https://labs.ripe.net/Members/ramakrishna_padmanabhan/reasons-dynamic-addresses-change) environment.

I hope to allocate some time in May for a blog post on `*.onion` scraping if you're still interested :-)

ansible/deploy-tor.yml
bassosimone: Thanks for the docs! Yeah, I figured it's a complex exercise!

PR #303, opened 2019-04-25
Tighten up firewall for ooni/orchestra database

darkk:
@hellais I assume you may want to tighten up firewall settings a bit on a DB box tracking OONI Probes.

I'm not rolling that out myself as I'm unsure if the list is still up-to-date. Seems, it is, but I may be missing something.

### DIFF
--- a/ansible/inventory
+++ b/ansible/inventory
@@ -139,8 +139,8 @@ datacollector.infra.ooni.io
 hkgmetadb.infra.ooni.io
 amsmetadb.ooni.nu
 ssdams.infra.ooni.io
-# db-1.proteus.test.ooni.io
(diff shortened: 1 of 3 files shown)


hellais:
This has been done as part of the infrastructure oonification process. Closing.

hellais:
Actually that is not entirely true, it has been done in the sense that there is a new database host running here: https://github.com/ooni/sysadmin/blob/master/ansible/templates/iptables.filter.part/ams-db-orchestra.ooni.nu which has good firewall settings, however we still need to migrate the old database to this new host with all the data.

That is tracked as part of: https://github.com/ooni/sysadmin/issues/383

PR #308, opened 2019-04-29
Drop crutch.py from pipeline

hellais:
As this is now done at the collector side: https://github.com/TheTorProject/ooni-backend/pull/116, this is no longer needed.

This should also allow measurements that are being uploaded with modern OONI Probe (but gathered with 2.0.0) to be re-uploaded if the user wants to.


darkk:
I'd also suggest to drop:
- `copy: src=crutch.py dest=/opt/crutch.py mode=0755`
- `ansible/roles/airflow/files/crutch.py`
- files from `/data/ooni/private/stash/`

hellais:
Pushed a fix for those.

I will look into deleting `/data/ooni/private/stash/` once I roll this out.

PR #320, opened 2019-05-30
mon.ooni.nu: add federico@

FedericoCeratto:
Related to #319


hellais:
I merged this into #319

PR #323, opened 2019-06-06
Deploy pipeline-shovel 20190605-0a2af2d4

FedericoCeratto:
https://github.com/ooni/pipeline/pull/198


hellais:
Once this is deployed we should also merge and deploy: https://github.com/ooni/api/pull/92

PR #326, opened 2019-06-21
Add analysis command to docker-trampoline and DAG

FedericoCeratto:
Minor improvements

### DIFF
--- a/ansible/roles/airflow/files/airflow-dags/analysis.py
+++ b/ansible/roles/airflow/files/airflow-dags/analysis.py
@@ -1,0 +1,17 @@
+# -*- coding: utf-8 -*-
+#
+# Use the existing shovel_jump.sh script to start the analysis script
+#
+from airflow import DAG
+from airflow.operators.bash_operator import BashOperator
+from datetime import datetime, timedelta
+
+dag = DAG(
+    dag_id="analysis",
+    schedule_interval=timedelta(days=1),
+    start_date=datetime(2019, 6, 21),
+    catchup=False,
+    default_args={"retries": 0},
+)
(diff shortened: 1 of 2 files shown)


ansible/roles/airflow/files/airflow-dags/analysis.py
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
+#
+# Use the existing shovel_jump.sh script to start the analysis script
+#
+from airflow import DAG
+from airflow.operators.bash_operator import BashOperator
+from datetime import datetime, timedelta
+
+dag = DAG(
+    dag_id="analysis_dag",
```
bassosimone: Minor: cursory looking at other DAGs seem to suggest we don't use the `_dag` prefix there. So, I'd recommend to avoid using the same prefix also here.

ansible/roles/airflow/files/docker-trampoline
```diff
@@ -227,6 +227,15 @@ case "$script" in
     exit 1 # if exec fails
     ;;
 
+    analysis)
+    volargs=""
+    volargs="${volargs} --volume=$public/analysis:/srv/analysis:rw"
+    volargs="${volargs} --tmpfs /tmp:rw,noexec,nosuid,size=128m"
+    rundocker="/usr/bin/docker run --interactive --rm --read-only --log-driver=journald"
+    $rundocker $volargs --name analysis openobservatory/pipeline-analysis:latest /srv/analysis.py
```
bassosimone: I am missing where `/srv/analysis.py` is

PR #334, opened 2019-07-15
Add comment on built-in docker init

darkk:
I'm cleaning up various notes and found a note on `dumb-init`. Actually, it's no longer needed, it's possible to replace it with `tini` that is shipped with modern docker versions.

I'm leaving it as a comment instead of separate ticket as there is no _need_ to drop `dumb-init`, it's just some cleanup.


hellais:
Thanks for this it's useful!

PR #335, opened 2019-07-22
Update orchestrate to 0.3.0

hellais:
So far I deployed 0.3.0 only on the testing infrastructure.

I think we should roll this out on the production endpoints too as it will increase testing coverage for people that have increased the maximum timeout and it has a small extra bandwidth cost.

The maximum response size will go up to 10x the current size, but it's still well under 100KB (77KB in the case of Iran which has a very big test list). Considering the fact that the web_connectivity test will likely consume much more bandwidth just to run the checks and do the upload I think it's reasonable to roll this out on production.

We may want to in the future to also do some smarter logic to impose a limit in the app by passing the `limit=` query parameter to the request.


hellais:
We agreed to deploying this on the production endpoint and see if anything breaks.

I am doing that now.

hellais:
This closes: https://github.com/ooni/probe/issues/860

PR #353, opened 2019-09-17
Upgrade dom0

hellais:
This is related to ensuring all our infrastructure is redeployable. 

In this PR I am going to address the issues of upgrading dom0-bootstrap to work with Debian 10 and also include fixes for: https://github.com/ooni/sysadmin/issues/289.


.travis.yml
```diff
@@ -2,7 +2,7 @@ language: python
 python:
   - "2.7"
 install:
-  - pip install ansible==2.4.2
+  - pip install ansible==2.8.3
```
bassosimone: I see in the README.md that we're forcing `dnspython==1.15.0`. Do we want to do that also here?

ansible/roles/dom0-defaults/tasks/main.yml
```diff
@@ -13,24 +13,22 @@
 
 - name: install some common packages
   apt:
-    name: "{{ item }}"
+    name:
+      - ca-certificates
+      - tmux
+      - screen
```
bassosimone: @hellais do you still prefer `screen` over `tmux`?

.travis.yml
hellais: That is probably a good idea too.

ansible/roles/passwd/tasks/main.yml
```diff
@@ -1,33 +0,0 @@
-# different system may have different places for bash and nologin
```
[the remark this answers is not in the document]
bassosimone: OK, thanks for explaining. I believe this is a good idea.

PR #361, opened 2019-09-30
Define all the necessary nginx endpoints for oonifying our services

hellais:
This implements what is described inside of: https://github.com/ooni/sysadmin/issues/359.

This is still work in progress, but I will use this to ask questions as I move forward with it.


ansible/roles/probe-services/templates/probe-services-nginx
```diff
@@ -0,0 +1,104 @@
...
+    # Should match:
+    # - /
+    # - /status
+    #
+    # The fact that it responds to / means that we may have to differentiate
+    # via the Host record.
+    # TODO We should check if clients will respect a suffix added to by the
+    # bouncer in the returned field, otherwise new clients should use another
+    # form
+    location ~^/web-connectivity/(status) {
```
hellais: @bassosimone do you know the answer to the above question for measurement-kit?

Basically if I return to a measurement-kit probe instead of:
```
      - {address: 'https://b.web-connectivity.th.ooni.io:443', type: https}
```

I say:

```
      - {address: 'https://b.web-connectivity.th.ooni.io:443/web-connectivity', type: https}
```

Will the client logic know to append the suffix web-connectivity or not?

hellais: Unfortunately it seems like this strategy will not work for legacy OONI Probe clients (see: https://github.com/ooni/probe-legacy/blob/master/ooni/backend_client.py#L52), so we will most likely still have to do `Host` header based routing.

bassosimone: Maybe we can see to improve the behaviour for new clients?

hellais: > Maybe we can see to improve the behaviour for new clients?

Yeah for sure. We probably want new clients to also be aware of the fact that they can speak any of the probe services APIs to any of the hosts (i.e. that the backends are oonified).

bassosimone: https://github.com/ooni/probe-engine/issues/63

hellais:
As soon as the CI converges I am going to merge this PR since it's deployed and so we don't have stale long running PRs on ooni/sysadmin.

PR #375, opened 2019-10-18
Oonification

hellais:
Fresh PR to work on Oonification of services


ansible/roles/ooni-measurements/tasks/main.yml
```diff
@@ -1,4 +1,16 @@
...
+    name: "oomsmweb"
+    state: present
+
+- name: create user
+  user:
+    name: "oomsmweb"
+    group: "oomsmweb"
+    comment: "OONI API user"
+    state: present
+
```
hellais: No idea why this diff is showing up, but it's already in master. It's likely due to the merge commit confusing github.

PR #376, opened 2019-10-23
Deploy ooni-explorer to ams-explorer.ooni.nu

sarathms:


sarathms:
Deployment fails with
```
fatal: [ams-explorer.ooni.nu]: FAILED! => {"changed": false, "checksum": "36e20eeb0894e52d7705390b27eea86e7e9d7c11", "msg": "Destination directory /etc/nginx/sites-enabled does not exist"}
```

Appears to be because installing `nginx` doesn't create `/etc/nginx/sites-available` automatically.

sarathms:
Deployment fails with
```
fatal: [ams-explorer.ooni.nu]: FAILED! => {"changed": false, "checksum": "36e20eeb0894e52d7705390b27eea86e7e9d7c11", "msg": "Destination directory /etc/nginx/sites-enabled does not exist"}
```

Appears to be because installing `nginx` doesn't create `/etc/nginx/sites-available` automatically.

hellais:
I think that is due to the fact that you were running the deployment script (based on what I see in #ooni-bots) with `-t explorer`:

```
sarath@debian runs ./play deploy-explorer.yml -t explorer on top of dirty 0d75a59 (master)
```

In order for that directly (as well as other necessary files) to exist you need to have first run the letsencrypt role.

Try re-running this with `./play deploy-explorer.yml` and it should work.

PR #378, opened 2019-10-25
Add TCP RTT metrics

hellais:
Missing is making the tcpmetrics into a daemon to continously write to
the file or setting up the systemd service to make it run periodically


hellais:
I merged this into oonification and will merge it into master from there.

Use that branch if you need to make any changes.

Closing this PR!

PR #380, opened 2019-10-29
Deploy fastpath

FedericoCeratto:
Push sshfeeder pubkey on collectors

(based on top of oonification)


ansible/roles/ooni-collector/tasks/main.yml
```diff
@@ -154,4 +154,4 @@
   authorized_key:
     user: sshfeeder
     state: present
-    key: "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAID3MZP0tYUzaZOZa8JhRIx3PiTXwXfwQdTPa/RwDmwcA"
+    key: " {{ sshfeeder_pubkey_b64encoded | b64decode }} "
```
hellais: I believe this will not work as the variable has been registered in the local scope of the `fastpath` role.

You can either do something with `set_fact`, however you make the two roles depend on each other.

Maybe it's better to just have the ssh key defined somewhere in the `group_vars` for the collector machines and have a step at the end of the `fastpath` role that checks if you are able to ssh into the machine properly and if not prompts you to update the value.

ansible/inventory
```diff
@@ -190,6 +190,9 @@ staticiforme.torproject.org       # tor LDAP
 [no_nodeexp:children]
 no_passwd # no passwd => no nodeexp user ;-)
 
+[fastpath]
```
hellais: Is there a reason to have it be part of the `fastpath` group?

ansible/roles/ooni-collector/tasks/main.yml
hellais: Maybe another way of doing it is like this: https://github.com/ooni/sysadmin/commit/2b8bc067a241e82e3c27c25eecfcb60dc9cb98dd

But I am also not a fan of this approach as it makes the playbook depend on the order in which you run the tasks.

I think having it as a hardcoded variable is probably the best approach.

FedericoCeratto: Variable removed; key updated.

ansible/roles/fastpath/tasks/main.yml
```diff
@@ -0,0 +1,37 @@
...
+# Fetch SSH pubkey and deploy it on the collector hosts
+#
+# Set the hostname of the collectors in the fastpath config file
+
+# Usage: place the .deb file in files/
+#        update the version number as needed
+
+- name: copy fastpath .deb package
+  copy:
+    src: files/fastpath_0.1_all.deb
```
hellais: Where can I obtain this file from?

FedericoCeratto:
Implements #394

ansible/inventory
FedericoCeratto: Should I use has_fastpath instead?

PR #393, opened 2019-11-22
currently long lines are cut off on irc

KwadroNaut:
Default is a message max of 512 bytes on IRC, not sure if this bot does the right thing, but according to its docs that should be it: https://node-irc.readthedocs.io/en/latest/API.html#irc.Client


hellais:
Thanks!

hellais:
This should fix: https://github.com/ooni/sysadmin/issues/342

PR #411, opened 2019-12-17
Deploy orchestra v0.3.1 and distribute psiphon config

bassosimone:


ansible/roles/probe-services/tasks/main.yml
```diff
@@ -31,6 +31,11 @@
     src: ooni-orchestrate.toml
     dest: "{{ orchestrate_data_dir }}/ooni-orchestrate.toml"
 
+- name: Write Psiphon config file
+  copy:
+    content: "{{ orchestra_psiphon_config_file_content }}"
```
hellais: You need to also defined this variable inside of the `defaults/main.yml` and probably set it to the `{{ CHANGE_ME }}` so that if you by mistake forget to define it in the scope of the target host, it will trigger an error.

If you don't define a default for it, when the role is getting compiled (which is before it's **evaluated** with the target host which is inside of the `psk_orchestration_prod` group) you will get an error.

ansible/group_vars/psk_orchestration/vars.yml
```diff
@@ -21,4 +21,5 @@ orchestra_apn_key_content_testing: '{{ vault_orchestra_apn_key_content_testing }
 
 orchestra_database_url_testing: 'postgres://{{ orchestra_database_username }}:{{ orchestra_database_password_testing }}@db-1.proteus.test.ooni.io:5432/{{ orchestra_database_name }}?sslmode=require'
 orchestra_database_url: 'postgres://{{ orchestra_database_username }}:{{ orchestra_database_password }}@db-1.proteus.ooni.io:5432/{{ orchestra_database_name }}?sslmode=require'
+orchestra_psiphon_config_file_content: '{{ vault_orchestra_psiphon_config_file_content }}'
```
hellais: This group `psk_orchestration` actually is not used for probe-services hosts: https://github.com/ooni/sysadmin/blob/master/ansible/inventory#L79.

I suggest you use the group `psk_orchestration_prod` and `psk_orchestration_test`

ansible/roles/probe-services/tasks/main.yml
bassosimone: 50a52b9

ansible/group_vars/psk_orchestration/vars.yml
bassosimone: fefba87

hellais:
This is now deployed in production. Merging.

PR #416, opened 2020-01-02
Drop references to brie.darkk.net.ru

### LINKED ISSUE #415: Drop brie.darkk.net.ru from monitoring

bassosimone:
Closes #415


ansible/deploy-cross-monitoring.yml
```diff
@@ -1,29 +0,0 @@
...
-  tasks:
-  # that's copy-paste from `prometheus` role with slightly different path
-  - name: Copy Exporter CA cert
-    copy:
-      src: exporter_ca.cert
-      dest: /srv/prometheus/ssl/exporter_ca.cert
-      owner: root
-      group: root
-      mode: 0644
-...
```
hellais: I think we should leave the cross-monitoring playbook as we may eventually want to do cross monitoring with another host.

bassosimone: 99% of that file was referring to the `brie` host though. After removing all the `brie` host related references, the only thing the file was doing was importing the ansible-version playbook.

hellais: As discussed in chat we should keep this as-is, so we can easily with a search and replace of the brie host, in theory, redeploy the brie host.

PR #420, opened 2020-01-16
Add very basic monitoring and restarting of the web-connectivity test…

hellais:
… helper service

This is an easier version of: https://github.com/ooni/backend/issues/274

### DIFF
--- a/ansible/roles/th-web-connectivity/templates/if-nok-restart.sh
+++ b/ansible/roles/th-web-connectivity/templates/if-nok-restart.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-if curl -m 5 -s {{ web_connectivity_ipv4 }}:{{ web_connectivity_port }}/status | grep '{"status": "ok"}' > /dev/null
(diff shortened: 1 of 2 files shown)


ansible/roles/th-web-connectivity/templates/if-nok-restart.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+if curl -m 5 -s {{ web_connectivity_ipv4 }}:{{ web_connectivity_port }}/status | grep '{"status": "ok"}' > /dev/null
```
bassosimone: do we want to add `-f` such that non `200` status codes are interpreted as bad?

hellais: I suppose we could do that, however the thing that matters the most is the fact that the response code matches the grep parameter.

Moreover the only case we care about in reality is the one in which curl times out, which is what happens when it enters the failed OOM state.

bassosimone: ok

hellais:
Thanks for the code @FedericoCeratto !

PR #424, opened 2020-02-10
Add latest default obfs4 bridge.

NullHypothesis:
The bridge exposes a port over both IPv4 and IPv6.

### DIFF
--- a/ansible/roles/probe-services/templates/tor_targets.json
+++ b/ansible/roles/probe-services/templates/tor_targets.json
@@ -300,5 +300,31 @@
       ]
     },
     "protocol": "obfs4"
+  },
+  "80390900eecf5e7820083775d66c8058d4cc2098840ec42cf1e2046f50552c9b": {
+    "address": "45.145.95.6:27015",
+    "fingerprint": "C5B7CD6946FF10C5B3E89691A7D3F2C122D2117C",
+    "params": {
+      "cert": [
+        "TD7PbUO0/0k6xYHMPW3vJxICfkMZNdkRrb63Zhl5j9dW3iRGiCx0A7mPhe5T2EDzQ35+Zw"
+      ],
+      "iat-mode": [
+        "0"
+      ]
+    },
+    "protocol": "obfs4"
+  },
+  "80390900eecf5e7820083775d66c8058d4cc2098840ec42cf1e2046f50552c9b": {
+    "address": "[2a0c:4d80:42:702::1]:27015",
+    "fingerprint": "C5B7CD6946FF10C5B3E89691A7D3F2C122D2117C",
+    "params": {
+      "cert": [
+        "TD7PbUO0/0k6xYHMPW3vJxICfkMZNdkRrb63Zhl5j9dW3iRGiCx0A7mPhe5T2EDzQ35+Zw"
+      ],
+      "iat-mode": [
+        "0"
+      ]
+    },
+    "protocol": "obfs4"
   }
 }
--- a/ansible/roles/probe-services/templates/tor_targets.json
+++ b/ansible/roles/probe-services/templates/tor_targets.json
@@ -119,7 +119,7 @@
     "fingerprint": "24E2F139121D4394C54B5BCC368B3B411857C413",
     "protocol": "dir_port"
   },
-  "e908d2aee266d911ca877b08b76d6dbc3ec9ad0c173c63eee59f3041d5ad54d4": {
+  "3fa772a44e07856b4c70e958b2f6dc8a29450a823509d5dbbf8b884e7fb5bb9d": {
     "address": "192.95.36.142:443",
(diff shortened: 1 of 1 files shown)


NullHypothesis:
Is the notion of the IPv6 address in `"address": "[2a0c:4d80:42:702::1]:27015",` correct?

ansible/roles/probe-services/templates/tor_targets.json
```diff
@@ -300,5 +300,31 @@
...
+        "TD7PbUO0/0k6xYHMPW3vJxICfkMZNdkRrb63Zhl5j9dW3iRGiCx0A7mPhe5T2EDzQ35+Zw"
+      ],
+      "iat-mode": [
+        "0"
+      ]
+    },
+    "protocol": "obfs4"
+  },
+  "80390900eecf5e7820083775d66c8058d4cc2098840ec42cf1e2046f50552c9b": {
+    "address": "[2a0c:4d80:42:702::1]:27015",
```
hellais: I am not sure we support ipv6 testing yet.

@bassosimone can you confirm this?

ansible/roles/probe-services/templates/tor_targets.json
```diff
@@ -300,5 +300,31 @@
       ]
     },
     "protocol": "obfs4"
+  },
+  "80390900eecf5e7820083775d66c8058d4cc2098840ec42cf1e2046f50552c9b": {
```
hellais: This is a duplicate key in the JSON object, I guess because the bridge is reachable on both ipv4 and ipv6 with the same fingerprint.

I think we need to think of how we go about testing ipv6 bridges as it's something we did not consider up until now.

An option could be of adding an additional key for the ipv6 address (`address_ipv6`?) and handle it in the testing logic.

@NullHypothesis given a bridge which is reachable on ipv6 and ipv4 can other parameters of the config (ex. the shared secret) vary depending on if you reach it via v4 or v6?

NullHypothesis: > @NullHypothesis given a bridge which is reachable on ipv6 and ipv4 can other parameters of the config (ex. the shared secret) vary depending on if you reach it via v4 or v6?

In this particular case that cannot happen because this bridge runs a single Tor process that spawned a single obfs4proxy process. Other parameters can vary however if somebody set up an IPv4/IPv6 bridge by running two separate Tor instances which spawn two separate obfs4proxy instances – but I believe you weren't asking about that.

ansible/roles/probe-services/templates/tor_targets.json
bassosimone: Yes, we do!

bassosimone: (@NullHypothesis also, the address format is OK)

ansible/roles/probe-services/templates/tor_targets.json
bassosimone: > I think we need to think of how we go about testing ipv6 bridges as it's something we did not consider up until now.

Append `/ipv4` or `/ipv6` to the fingerprint and then hash that? (Or append `/ipv4` or `/ipv6` to the hash, which maybe is even more clear?)

NullHypothesis: > > I think we need to think of how we go about testing ipv6 bridges as it's something we did not consider up until now.
> 
> Append `/ipv4` or `/ipv6` to the fingerprint and then hash that? (Or append `/ipv4` or `/ipv6` to the hash, which maybe is even more clear?)

I'm happy to do that but why not make the key be addr:port as it is done for the directory authorities?

hellais: > I'm happy to do that but why not make the key be addr:port as it is done for the directory authorities?

The reason for that is that we would like to eventually support testing also private bridges which come from bridge-db and if we are using the IP and port as key, we would be disclosing the address of the bridge in the result.

As a matter of fact, you can use whatever you prefer as a key for the testing target, we figured using a hash of the fingerprint was a sane approach that would work well also in the case of private bridges.

If you would prefer we use a different key, we can do that too.

NullHypothesis: That makes sense, thanks for elaborating! I'll revise my pull request.

NullHypothesis: In 13d7c81, I derived a bridge's key from the concatenation of its IP:port tuple and its fingerprint. This should be future-proof and even works in a world where a Tor bridge exposes multiple IPv4/IPv6 endpoints using a single identity.

NullHypothesis:
I created https://github.com/ooni/spec/pull/177 to explain how we now choose the key.

hellais:
Thanks for putting this together! I am going to squash, merge and deploy!

PR #437, opened 2020-03-19
Switch to using hkgmetadb on ams-api

hellais:


FedericoCeratto:
Related to https://github.com/ooni/backend/issues/371

PR #439, opened 2020-03-24
Add API robots.txt

FedericoCeratto:


ansible/roles/ooni-measurements/templates/ngx-oomsm-web
```diff
@@ -49,6 +49,11 @@ server {
 
     {{ c.location_letsencrypt() }}
 
+    location = /robots.txt {
+       add_header Content-Type text/plain;
+       return 200 "User-agent: *\nDisallow: /\nAllow: /docs";
```
hellais: The API docs are in: https://api.ooni.io/api/

Also we should allow scraping the index of the API.

hellais: I would say we should make this be like follows:

```
User-agent: *
Allow: /
Disallow: /api/v1/
Disallow: /api/_/
Disallow: /files/
Disallow: /stats/
```

PR #446, opened 2020-06-05
Add haproxy and Prio

FedericoCeratto:


ansible/roles/probe-services/templates/probe-services-nginx
```diff
@@ -141,5 +157,19 @@ server {
         return 200 "{{ probe_services_domain }}";
     }
 
+    # Haproxy on localhost dashboard - see roles/haproxy/README.adoc
+    # - /haproxy_stats
```
hellais, line 161: Is it OK for this endpoint to be publicly visible or might it present security concerns?

PR #448, opened 2020-06-23
Fix letsencrypt role

FedericoCeratto:
https://github.com/ooni/backend/issues/377


ansible/roles/letsencrypt/tasks/main.yml
```diff
@@ -33,46 +18,32 @@
...
   command: >
     certbot certonly
     --webroot --webroot-path /var/www/letsencrypt
     --preferred-challenges http
     --noninteractive --text --agree-tos
     --email {{ letsencrypt_admin_email }}
     --domains {{ letsencrypt_domains | join(',') }}
-    {% if expand is defined %}--expand{% endif %}
+    --expand
+    --keep
```
hellais, line 46: Are you sure we should not be using the `--nginx` option for nginx?

ansible/roles/letsencrypt/tasks/main.yml
```diff
@@ -33,46 +18,32 @@
...
-  ignore_errors: yes
-  register: acme_challenge_present
-
-- name: Add letsencrypt vhost
+- name: Add letsencrypt dedicated sites-enabled file
   template:
     src: templates/letsencrypt-http
     dest: /etc/nginx/sites-enabled/letsencrypt-http
+    mode: 0644
+    owner: root
```
hellais, line 26: The pattern we were using up until now was that of writing directly inside of the default nginx config by calling the template functions: https://github.com/ooni/sysadmin/blob/master/ansible/templates/common.j2.

The reason to do so is that `letsencrypt-http` will be mapped to port 80, meaning that we are not able to serve the website on port 80 or do a redirect to port 443 in cases in which that is needed: https://github.com/ooni/sysadmin/blob/8e9c599b5f845a2914ed0c52c9191e5f40a17eb6/ansible/roles/letsencrypt/templates/letsencrypt-http

ansible/roles/letsencrypt/templates/letsencrypt-http
```diff
@@ -1,12 +1,18 @@
...
+    server_name _;
 
+    # Serve ACME challenge from disk
     location /.well-known/acme-challenge {
         root /var/www/letsencrypt;
         try_files $uri $uri/ =404;
     }
+
+    # Redirect everything else to port 443 regardless of domain
+    return 301 https://$host$request_uri;
```
hellais, line 17: Yes this is a good idea. 

In case you haven't done so already, I suggest:
*  Checking if there are some services which require nginx to serve the service on port 80 and not enforce TLS (I don't think that is the case, but it's worth checking)
* Grepping for `c.location_letsencrypt()` and removing that template from there
* Dropping from `ansible/templates/common.j2` all functions which are no longer needed

ansible/roles/letsencrypt/tasks/main.yml
hellais: This questions/comment still is relevant I think.

FedericoCeratto: Certbot is writing the challenge in a directory and calling the post hook  " --post-hook '/usr/sbin/service nginx reload'" during cert updates.
It's not using the Nginx plugin anymore and it's unaware of Nginx in itself.

hellais:
Another thing that might be helpful is adding some documentation on how one would go about integrating the letsencrypt role into a service which requires a TLS certificate. If there is any special thing they should be aware of (for example the fact that it will automatically do a redirect from port 80 to 443) or any other integration work that should be done.

hellais:
In any case this PR is good to merge for me!

ansible/roles/letsencrypt/templates/letsencrypt-http
FedericoCeratto: This PR is only for the letsencrypt role. Removing location_letsencrypt() in other roles should be done when we'll migrate the other roles to the new letsencrypt setup in future PRs.

FedericoCeratto:
Added README to clarify usage

PR #449, opened 2020-06-26
Postgresql 11 role

FedericoCeratto:


ansible/deploy-postgres-11.yml
```diff
@@ -0,0 +1,43 @@
...
+  tags: postgresql11
+  roles:
+    - role: postgresql11
+
+
+      # etckeeper
+      # netdata 
+      # firewall
+
+#- name: install prometheus-postgres-exporter
```
hellais, line 18: I think you need to nest tasks inside of the task key instead of being at the role level. It's probably better though to put these inside a dedicated role.

ansible/roles/base-buster/tasks/main.yml
```diff
@@ -0,0 +1,49 @@
...
+      - nullmailer
+      - prometheus-node-exporter
+      - rsync
+      - strace
+      - tcpdump
+      - tmux
+
+- name: Clean cache
+  apt:
+    autoclean: yes
```
hellais, line 49: Do you think we should be adding in here also the creation of the accounts, like what dom0 is doing?

ansible/roles/postgresql11/tasks/main.yml
```diff
@@ -0,0 +1,41 @@
...
+  lvg:
+    vg: vg.metadb
+    pvs:
+      - /dev/xvdb1
+      - /dev/xvdc1
+      - /dev/xvdd1
+      - /dev/xvde1
+      - /dev/xvdf1
+      - /dev/xvdg1
+      - /dev/xvdh1
```
hellais, line 12: Maybe, since these are about the specific disk setup of the particular ams-pg host, these should go outside of the postgresql11 role?

ansible/deploy-postgres-11.yml
FedericoCeratto: Yes, the comments were just temporary TODOs. Removed

ansible/roles/base-buster/tasks/main.yml
FedericoCeratto: The bootstrap is quite different from dom0, maybe we can deploy new nodes with these new roles over the next months before starting to backport stuff (due to the risk of breaking live nodes).

ansible/roles/postgresql11/tasks/main.yml
FedericoCeratto: Active and standby DB hosts are pretty critical. I think it's safer to dedicate a role for a specific host (instead of having variables and conditionals that can cause unexpected changes on a host A while we are changing host B).

hellais: Makes sense, perhaps then the role name should contain the active/standby role in it?

ansible/roles/base-buster/tasks/main.yml
hellais: So currently to deploy this, you would still have to run the dom0 bootstrap playbook on it?

ansible/roles/base-buster/tasks/main.yml
```diff
@@ -0,0 +1,125 @@
...
+    create: yes
+    block: |
+      # Managed by ansible, see roles/base-buster/tasks/main.yml
+      update_every: 5
+      local:
+        command: 'chronyc -n tracking'
+
+- name: configure netdata nginx
+  blockinfile:
+    path: /etc/netdata/python.d/chrony.conf
```
hellais, line 102: Is this a typo, or is it normal that it’s writing to the same file?

ansible/roles/base-buster/tasks/main.yml
```diff
@@ -0,0 +1,125 @@
...
+    create: yes
+    block: |
+      add rule inet filter input ip saddr {{ lookup('dig', 'prometheus.infra.ooni.io/A') }} tcp dport 19999 counter accept comment "netdata.service"
+
+- name: reload nftables service
+  systemd:
+    name: nftables.service
+    state: reloaded
+
+- name: configure netdata.service
```
hellais, line 76: Is there a reason to use blockinfile instead of file or template?

ansible/roles/postgresql11/meta/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+dependencies:
```
hellais, line 2: So the key here is that the role itself will specify the dependencies for other roles, correct?
I think this a reasonable approach, though we should probably document this architecture/pattern somewhere as it might not be immediately obvious this is how roles should be written from now on.

ansible/roles/prometheus/templates/prometheus.yml.j2
```diff
@@ -51,7 +51,7 @@ scrape_configs:
       module: [{{ bbjob.module }}]
     static_configs:
       - targets:
-{% for target in bbjob.targets %}
+{% for target in (bbjob.targets|sort) %}
```
hellais, line 54: Was the sort added to make the diff more deterministic? I am surprised that ansible did not do it automatically.

ansible/roles/prometheus/templates/prometheus.yml.j2
```diff
@@ -151,9 +151,30 @@ scrape_configs:
...
+
+  - job_name: 'raw postgres-exporter'
+    scrape_interval: 5s
+    scheme: https
+    metrics_path: /api/v1/allmetrics
+    params:
+      format: [prometheus]
+    static_configs:
+      - targets:
+        - ams-pg.ooni.nu:9187
```
hellais, line 177: What is the difference between these two netadata endpoints and the ones that have_netdata?

ansible/roles/base-buster/meta/main.yml
```diff
@@ -0,0 +1,8 @@
+---
+dependencies:
+  - role: track_etc_directory
```
hellais, line 3: Maybe this is a bit too much indirection? Can we put the content of this in the role directly for the time being and abstract it when N > 1 roles need to depend on this and not the base one?

ansible/roles/nginx-buster/templates/nginx.conf
```diff
@@ -0,0 +1,101 @@
+
+# Managed by ansible
+# roles/nginx-buster/templates/nginx.conf
```
hellais, line 3: Where is this role being included?

ansible/roles/base-buster/meta/main.yml
FedericoCeratto: Moved to base-buster and dom0-bootstrap

ansible/roles/base-buster/tasks/main.yml
FedericoCeratto: Better readability: we can scan through the task file and see what is being done.

ansible/roles/base-buster/tasks/main.yml
FedericoCeratto: Updated to create users

ansible/roles/prometheus/templates/prometheus.yml.j2
FedericoCeratto: Yes the first, no to the latter: the diff was random every time

PR #483, opened 2021-10-01
Update Namecheap DNS fetching

FedericoCeratto:


hellais (commented):
I suggest doing the following:
- Remove the .txt and .xml zone files from the repo
- Delete ../ext/hosts as it depends on the .xml files which we no longer support generating
- Move the svg generator out of the ansible script (so we don't have an additional command dependency for running the update of the DNS records)
- Commit a version of the svg charts to the branch

Once that's done, it's good to go.

hellais (approved):
LGTM

PR #495, opened 2022-05-16
Tor targets

hellais:
Upon manual inspection I noticed that 2 things change (it's hard to see
in the diff because we weren't previously sorting keys):

1. The removal of the smallerRichard bridge as requested by @meskio in https://github.com/ooni/sysadmin/pull/493
2. The id of the bridge 51.222.13.177:80 changed from
fa69b2ee7a7c8975af2452ecd566f67a6459d397a4cefc30be86a670675cdc23 to
662218447d396b9d4f01b585457d267735601fedbeb9a19b86b942f238fe4e7b. The
new ID is correct and it's unclear how this ID was determined in the
original PR: https://github.com/ooni/sysadmin/pull/463.

The second point might create data analysis issues in the future, but I
don't really see a way around it except remembering about this
inconsistency so we can manually remap the in the future.

### DIFF
--- a/ansible/roles/probe-services/templates/tor_targets.json
+++ b/ansible/roles/probe-services/templates/tor_targets.json
@@ -1,123 +1,100 @@
 {
   "128.31.0.39:9101": {
     "address": "128.31.0.39:9101",
-    "name": "moria1",
     "fingerprint": "9695DFC35FFEB861329B9F1AB04C46397020CE31",
+    "name": "moria1",
     "protocol": "or_port_dirauth"
   },
   "128.31.0.39:9131": {
     "address": "128.31.0.39:9131",
-    "name": "moria1",
     "fingerprint": "9695DFC35FFEB861329B9F1AB04C46397020CE31",
-    "protocol": "dir_port"
-  },
-  "86.59.21.38:443": {
-    "address": "86.59.21.38:443",
-    "name": "tor26",
-    "fingerprint": "847B1F850344D7876491A54892F904934E4EB85D",
-    "protocol": "or_port_dirauth"
-  },
-  "86.59.21.38:80": {
-    "address": "86.59.21.38:80",
-    "name": "tor26",
-    "fingerprint": "847B1F850344D7876491A54892F904934E4EB85D",
-    "protocol": "dir_port"
-  },
-  "45.66.33.45:443": {
-    "address": "45.66.33.45:443",
-    "name": "dizum",
-    "fingerprint": "7EA6EAD6FD83083C538F44038BBFA077587DD755",
-    "protocol": "or_port_dirauth"
-  },
-  "45.66.33.45:80": {
-    "address": "45.66.33.45:80",
-    "name": "dizum",
-    "fingerprint": "7EA6EAD6FD83083C538F44038BBFA077587DD755",
-    "protocol": "dir_port"
-  },
-  "66.111.2.131:9001": {
-    "address": "66.111.2.131:9001",
-    "name": "Serge",
-    "fingerprint": "BA44A889E64B93FAA2B114E02C2A279A8555C533",
-    "protocol": "or_port_dirauth"
-  },
-  "66.111.2.131:9030": {
-    "address": "66.111.2.131:9030",
-    "name": "Serge",
-    "fingerprint": "BA44A889E64B93FAA2B114E02C2A279A8555C533",
+    "name": "moria1",
     "protocol": "dir_port"
   },
   "131.188.40.189:443": {
     "address": "131.188.40.189:443",
-    "name": "gabelmoo",
     "fingerprint": "F2044413DAC2E02E3D6BCF4735A19BCA1DE97281",
+    "name": "gabelmoo",
     "protocol": "or_port_dirauth"
   },
   "131.188.40.189:80": {
     "address": "131.188.40.189:80",
-    "name": "gabelmoo",
     "fingerprint": "F2044413DAC2E02E3D6BCF4735A19BCA1DE97281",
-    "protocol": "dir_port"
-  },
-  "193.23.244.244:443": {
-    "address": "193.23.244.244:443",
-    "name": "dannenberg",
-    "fingerprint": "7BE683E65D48141321C5ED92F075C55364AC7123",
-    "protocol": "or_port_dirauth"
-  },
-  "193.23.244.244:80": {
-    "address": "193.23.244.244:80",
-    "name": "dannenberg",
-    "fingerprint": "7BE683E65D48141321C5ED92F075C55364AC7123",
-    "protocol": "dir_port"
-  },
-  "171.25.193.9:80": {
-    "address": "171.25.193.9:80",
-    "name": "maatuska",
-    "fingerprint": "BD6A829255CB08E66FBE7D3748363586E46B3810",
-    "protocol": "or_port_dirauth"
-  },
-  "171.25.193.9:443": {
-    "address": "171.25.193.9:443",
-    "name": "maatuska",
-    "fingerprint": "BD6A829255CB08E66FBE7D3748363586E46B3810",
+    "name": "gabelmoo",
     "protocol": "dir_port"
   },
   "154.35.175.225:443": {
     "address": "154.35.175.225:443",
-    "name": "Faravahar",
     "fingerprint": "CF6D0AAFB385BE71B8E111FC5CFF4B47923733BC",
+    "name": "Faravahar",
     "protocol": "or_port_dirauth"
   },
   "154.35.175.225:80": {
     "address": "154.35.175.225:80",
-    "name": "Faravahar",
     "fingerprint": "CF6D0AAFB385BE71B8E111FC5CFF4B47923733BC",
+    "name": "Faravahar",
+    "protocol": "dir_port"
+  },
+  "171.25.193.9:443": {
+    "address": "171.25.193.9:443",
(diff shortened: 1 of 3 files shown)


bassosimone:
> The second point might create data analysis issues in the future, but I
> don't really see a way around it except remembering about this
> inconsistency so we can manually remap the in the future.

Wouldn't the new fingerprint make the bridge appear as a new bridge? (Which is the primary key we use for selecting bridges for the purpose of testing?)

Which would otherwise be the best place to document this oddity in your opinion?

scripts/make_tor_targets.py
```diff
@@ -0,0 +1,114 @@
...
+    return bridge_id, bridge
+
+def get_bridges():
+    with urlopen(BRIDGES_URL) as resp:
+        j = json.loads(resp.read())
+
+    bridges = {}
+    for b in j["obfs4"]:
+        bridge_id, bd = parse_bridge_line(b)
+        assert bridge_id not in bridges
```
bassosimone, line 37: You should probably explain with a comment why the policy is to fail the script in case of duplicates. (Other policies may be possible, such as using the first or the last entry, but here you chose explicitly to fail the script if that happens, which likely has some background reasoning. Because it may be that you're not the only future person running this script, I was wondering whether it would be best to explain the policy with a comment and provide additional instructions in the same comment regarding what the operator should do in this case.)

scripts/make_tor_targets.py
```diff
@@ -0,0 +1,114 @@
...
+        bridges[bridge_id] = bd
+    return bridges
+
+def parse_dirauth(line : str) -> dict:
```
bassosimone, line 41: As before, it would be beneficial to add an example we can publicly share as a comment, such that whoever is reading the code in the future could more easily and immediately view that the code is correct.

```suggestion
def parse_dirauth(line : str) -> dict:
    # Example: tor26 orport=443 v3ident=14C131DFC5C6F93646BE72FA1401C02A8DF2E8B4 ipv6=[2001:858:2:2:aabb:0:563b:1526]:443 86.59.21.38:80 847B 1F85 0344 D787 6491 A548 92F9 0493 4E4E B85D
```

Again, the example is taken from the public API and so I'm comfortable adding it.

scripts/make_tor_targets.py
```diff
@@ -0,0 +1,114 @@
...
+    print(f"written {path} file")
+
+def main():
+    bridges = get_bridges()
+    dir_auths = get_dirauths()
+
+    res = {}
+    res.update(dir_auths)
+    res.update(bridges)
+
```
bassosimone, line 109: So, this is probably the right place where to document why we have two files using a comment.

scripts/make_tor_targets.py
```diff
@@ -0,0 +1,114 @@
...
+    return bridges
+
+def parse_dirauth(line : str) -> dict:
+    da = {}
+    parts = line.split(" ")
+    da["name"] = parts[0]
+    assert parts[1].startswith("orport=")
+    da["or_port"] = parts[1].lstrip("orport=")
+    da["dir_address"] = parts[-11]
+    da["fingerprint"] = "".join(parts[-10:])
```
bassosimone, line 48:
```suggestion
    da["fingerprint"] = "".join(parts[-10:])  # ditto
```

scripts/make_tor_targets.py
```diff
@@ -0,0 +1,114 @@
...
+        bridges[bridge_id] = bd
+    return bridges
+
+def parse_dirauth(line : str) -> dict:
+    da = {}
+    parts = line.split(" ")
+    da["name"] = parts[0]
+    assert parts[1].startswith("orport=")
+    da["or_port"] = parts[1].lstrip("orport=")
+    da["dir_address"] = parts[-11]
```
bassosimone, line 47:
```suggestion
    # TODO(hellais, bassosimone): This parsing algorithm ignores the IPv6 address.
    da["dir_address"] = parts[-11]  # Note: the fingerprint consists of 10 elements
```

scripts/make_tor_targets.py
```diff
@@ -0,0 +1,114 @@
...
+
+        line = line.strip().lstrip('"').rstrip('"')
+        current_line += line
+        if is_done:
+            da_lines.append(current_line)
+
+    for line in da_lines:
+        da = parse_dirauth(line)
+
+        or_address = da["dir_address"].split(":")[0] + ":" + da["or_port"]
```
bassosimone, line 77:
```suggestion
        # Note: this algorithm only manipulates IPv4 addresses
        or_address = da["dir_address"].split(":")[0] + ":" + da["or_port"]
```

Honestly here I would be happier to use correct parsing of endpoints rather than splitting by `:` because that would break when using IPv6. I would recommend borrowing more robust code to perform this kind of splitting from here: https://github.com/bassosimone/websteps-illustrated/blob/ef7a88b15ac5d13fb95525551e9cb078853cfa7b/python/websteps.py#L325

scripts/make_tor_targets.py
```diff
@@ -0,0 +1,114 @@
...
+        assert or_address not in dir_auths
+        dir_auths[or_address] = {
+            "address": or_address,
+            "name": da["name"],
+            "fingerprint": da["fingerprint"],
+            "protocol": "or_port_dirauth"
+        }
+
+        assert da["dir_address"] not in dir_auths
```
bassosimone, line 86: Same as above, maybe mention what an operator should do in case of error?

scripts/make_tor_targets.py
```diff
@@ -0,0 +1,114 @@
+import json
```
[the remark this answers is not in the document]
bassosimone: Also, it would be probably worth mentioning in some readme/manual for deployment what one needs to do when their goal is to update the list of tor bridges we use. (Or, is there some plan to perform this kind of update automatically when we're deploying? Would that be safe and desirable?)

bassosimone (commented):
Mostly LGTM but there are some talking points that I think we should discuss before proceeding with merging. My main concern here has been to double check the algorithms and suggest ways to either improve the code or document how we could improve the code.

Thank you for moving this work forward ❤️ 🥳 🦜

ansible/roles/probe-services/templates/tor_targets.json
```diff
@@ -1,124 +1,101 @@
 {
```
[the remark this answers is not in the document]
hellais: I don't know the answer to this. I updated both because I wasn't sure which one is actually used by the backend to provide data to probes.
@FedericoCeratto should know the answer to this.

If one of the two roles is deprecated, I think we should delete it.

scripts/make_tor_targets.py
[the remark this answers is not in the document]
hellais: In an ideal world this would be done automatically and periodically on the deployed host. Since I wasn't sure if we wanted to do this I implemented it for the time being as something that somebody has to manually run.

If we are to require manually running it, we should document it's usage in the backend docs, though I am not sure what is the correct place to do so.

We currently have backend related documentation in the following places:
- https://github.com/ooni/backend/wiki
- https://ooni.github.io/pipeline/
- https://github.com/ooni/pipeline#ooni-backend

It's unclear to me which one is the latest and where I should be putting this documentation.

cc @FedericoCeratto

scripts/make_tor_targets.py
```diff
@@ -0,0 +1,114 @@
...
+import hashlib
+
+from urllib.request import urlopen
+
+DIRAUTH_URL = "https://gitweb.torproject.org/tor.git/plain/src/app/config/auth_dirs.inc"
+BRIDGES_URL = "https://bridges.torproject.org/moat/circumvention/builtin"
+
+def parse_params(parts : list[str]) -> dict[str, str]:
```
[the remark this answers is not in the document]
hellais: Ah yeah you are right, I made changes to the function and forgot to update the typing (if only python did this automatically when you run the script without needing other tooling...).

bassosimone:
> > Wouldn't the new fingerprint make the bridge appear as a new bridge? (Which is the primary key we use for selecting bridges for the purpose of testing?)
> 
> Yeah basically it will mean we won't be able to easily link measurements for this bridge from the past with the future. I don't think it's a huge deal, but it's worth taking note of it.
> 
> > Which would otherwise be the best place to document this oddity in your opinion?
> 
> I think we could maybe add it as a note to the tor ooni/spec document?

Sounds good!

scripts/make_tor_targets.py
hellais: It's failing because the data format is a dict, so if there is a duplicate key, you will end up overwriting the previous entry without noticing.

I agree this should be added as a comment. I will do so.

scripts/make_tor_targets.py
FedericoCeratto, line 1: This script should go in the API package with its dedicated .timer and .service 
With logging and metrics similar to the API so that we can monitor the run and alarm when needed.
The 2 .json files can be removed from here. The API runner would start without .json files and download the first copy.

FedericoCeratto (changes requested):

ansible/roles/probe-services/templates/tor_targets.json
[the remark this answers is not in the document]
FedericoCeratto: The "probe-services" role is legacy and was used for the backend hosts on Greenhost like hkg-ps.ooni.nu 
The new backend uses "ooni-backend"

PR #506, opened 2023-12-20
fix(ooni-backend/haproxy): use path_beg to route ACME requests

bassosimone:
This diff fixes haproxy.cfg for the ooni-backend role such that we use the path_beg filter rather than using path for routing ACME requests.

The issue with using path rather than path_beg is that path seems to be an exact filter while path_beg matches the prefix.

See https://www.haproxy.com/blog/path-based-routing-with-haproxy for an introduction about using path and path_beg.


bassosimone:
I deployed it using `./play deploy-backend.yml -l ams-pg-test.ooni.org --diff -t webserv`.

We can self merge this PR now.

PR #509, opened 2023-12-22
Update test helper Nginx

FedericoCeratto:
Fix resolution, logging, ipaddr anonimization

### DIFF
--- a/ansible/roles/ooni-backend/templates/rotation_nginx_conf
+++ b/ansible/roles/ooni-backend/templates/rotation_nginx_conf
@@ -1,13 +1,36 @@
 # Managed by ansible, see roles/ooni-backend/tasks/main.yml
 # and roles/ooni-backend/templates/rotation_nginx_conf
 # Deployed by rotation tool to the test-helper hosts
+
+# Use 2-level cache using 5GB on disk
 proxy_cache_path /var/cache/nginx levels=1:2 keys_zone=thcache:100M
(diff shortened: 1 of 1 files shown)


ansible/roles/ooni-backend/templates/rotation_nginx_conf
```diff
@@ -1,13 +1,36 @@
...
+map $remote_addr $remote_addr_anon {
+  ~(?P<ip>\d+\.\d+\.\d+)\.    $ip.0;
+  ~(?P<ip>[^:]+:[^:]+):       $ip::;
+  default                     0.0.0.0;
+}
+
+# anonymize forwarded ipaddr
+map $http_x_forwarded_for $remote_fwd_anon {
+  ~(?P<ip>\d+\.\d+\.\d+)\.    $ip.0;
+  ~(?P<ip>[^:]+:[^:]+):       $ip::;
```
bassosimone, line 19: See https://stackoverflow.com/questions/6098517/anonymizing-ipv6-addresses for a conversation about how much one should strip of IPv6 addresses to make sense of them. So far, we're erring on the stripping more side but we may want to revisit this choice at a later time if stripping too much makes it impossible to make sense of IPv6 addresses in the logs (I am thinking, in particular, about attacks to mitigate).

bassosimone (approved):
🐳

PR #511, opened 2024-02-02
Fix debops

hellais:
Summary of changes:
* Ignore rebuilding of debian packages that have been deleted
* Add `--dry-run` CLI option to make it easier to locally test

As noted inside of https://github.com/ooni/backend/pull/798, debops-ci will attempt to build a package even when the package has been deleted.

This happens because it will rebuild any package that it detects by running:

```
git diff --name-only {commit}
```

and filtering by those names that match the string `debian/changelog`.

However, since it's not filtering by the status, which could also be that the `debian/changelog` had been deleted, it thinks it should rebuild also a deleted package.

In this patch I fix that, by ignoring any paths that have the `D` (deleted) status.

I also add support for a `--dry-run` command line flag, allowing you to better debug what's going on in non-debian systems.

### DIFF
--- a/tools/debops-ci
+++ b/tools/debops-ci
@@ -110,8 +110,11 @@ def detect_changed_packages() -> List[Path]:
     changes = run2(f"git diff --name-only {commit}")
     pkgs = set()
     for cs in changes.splitlines():
-        c = Path(cs)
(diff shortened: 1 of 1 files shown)


bassosimone (approved):
🐳 💯

PR #524, opened 2025-02-19
refactor: disable orchestrate and point to new ooniprobe ecs service

DecFox:
This diff points the backend-fsn host to the new ooniprobe service and also gets rid of the orchestra proxy which can now be served directly from backend-fsn

### DIFF
--- a/ansible/roles/ooni-backend/templates/nginx-api-fsn.conf
+++ b/ansible/roles/ooni-backend/templates/nginx-api-fsn.conf
@@ -82,7 +82,7 @@ server {
       proxy_set_header X-Forwarded-Proto $scheme;
       proxy_read_timeout 900;
 
-      proxy_pass https://registry.ooni.io:443;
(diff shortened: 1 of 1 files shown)


ansible/roles/ooni-backend/templates/nginx-api-fsn.conf
```diff
@@ -111,20 +111,20 @@ server {
...
-      proxy_set_header X-Forwarded-Proto $scheme;
-      proxy_read_timeout 900;
-
-      proxy_pass https://orchestrate.ooni.io:443;
-      add_header Strict-Transport-Security "max-age=31536000; includeSubDomains" always;
-      add_header X-Frame-Options DENY always;
-      add_header X-Content-Type-Options nosniff always;
-
-  }
+  # location ~^/api/v1/(test-list|urls) {
```
hellais, line 114: Maybe let's just add a comment above this to mention we are serving this from ooniapi monolith

